PR #1, opened 2019-02-05
Graphite Export Plugin

dokshina:
Added graphite export plugin and it's usage example.
Described how to create custom export plugin.

Renamed metrics.details.prometheus -> metrics.details.


README.md
```diff
@@ -141,6 +141,7 @@ metrics_client = require('metrics')
 
 #### `client_obj.collect()`
   Returns concatenation of `obj:collect()` across all collectors created.
+  See above `counter_obj:collect()` for details.
```
ASverdlov: Why only `counter_obj`?

metrics/plugins/graphite/init.lua
```diff
@@ -0,0 +1,79 @@
...
+local function graphite_worker(opts)
+    fiber.name('metrics_graphite_worker')
+
+    while true do
+        local observations = metrics.collect()
+        for _, obs in ipairs(observations) do
+            local data = format_observation(opts.prefix, obs)
+            local numbytes = opts.sock:sendto(opts.host, opts.port, data)
+            if numbytes == nil then
+                log.info('Error while sending to host %s port %s data %s',
```
ASverdlov: This is an error.
Maybe better use`log.error()` then?

dokshina: Yes, you are right

README.md
dokshina: Because it is the same as `gauge_obj:collect()`.
I think it is better to move `client_obj.collect()` section to the top (after `metrics_client.connect()`), add there `data` description, (and maybe rename `data` -> `observation`), and in `{counter,gauge,hist}:collect()` sections just reference to   `client_obj.collect()`.

ASverdlov: Good idea, agreed. We need to:
1) introduce `observation` object,
2) outline that is same for both `counter` and `gauge` object and
3) refer to its section in all `:collect()` calls.

ASverdlov:
Please consider `Graphite` histogram native format when exporting.

We probably need to adapt it. Internally we store histograms in `Prometheus`-like format.

ASverdlov:
Decided to merge this. We'll handle `graphite` histograms in a separate PR.

PR #2, opened 2019-03-12
RPM Spec

dokshina:
Added RPM spec and Makefile to easy build RPMs.


rpm/tarantool-metrics.spec
```diff
@@ -0,0 +1,41 @@
...
+%define luapkgdir %{_datadir}/tarantool
+%define br_luapkgdir %{buildroot}%{luapkgdir}
+
+%prep
+%setup -q -n %{name}-%{version}
+
+%install
+mkdir -p %{br_luapkgdir}
+cp -av metrics %{br_luapkgdir}
+ls %{br_luapkgdir}
```
Totktonada: I guess it is not needed here.

Totktonada:
This RPM is not deployed to our packagecloud repos?

PR #3, opened 2019-03-28
Add Prometheus Exporter

ASverdlov:
Implements Prometheus Exporter for `tarantool/metrics`.

New public function `metrics.collectors()` was created in order to export histograms correctly. This is because Prometheus HTTP payload requires additional metainfo on collector types in addition to raw observations.

`metrics.collectors()` returns a set of collectors.

[tarantool/prometheus](https://github.com/tarantool/prometheus) was used as a reference implementation.

### DIFF
--- a/metrics/details/init.lua
+++ b/metrics/details/init.lua
@@ -37,11 +37,13 @@ function Registry:unregister(collector)
     end
 end
 
-function Registry:collect()
+function Registry:invoke_callbacks()
     for _, registered_callback in ipairs(self.callbacks) do
         registered_callback()
     end
+end
 
+function Registry:collect()
     local result = {}
     for _, collector in pairs(self.collectors) do
         for _, obs in ipairs(collector:collect()) do
--- a/metrics/details/init.lua
+++ b/metrics/details/init.lua
@@ -14,21 +14,26 @@ Registry.__index = Registry
 function Registry.new()
     local obj = {}
     setmetatable(obj, Registry)
+
     obj.collectors = {}
     obj.callbacks = {}
     return obj
 end
 
 function Registry:register(collector)
-    if self.collectors[collector.name] ~= nil then
-        return self.collectors[collector.name]
+    for _, c in ipairs(self.collectors) do
+        if c.name == collector.name and c.collector == collector.collector then
+            return
+        end
     end
-    self.collectors[collector.name] = collector
+    table.insert(self.collectors, collector)
 end
 
 function Registry:unregister(collector)
-    if self.collectors[collector.name] ~= nil then
-        table.remove(self.collectors, collector.name)
+    for i, c in ipairs(self.collectors) do
+        if c.name == collector.name and c.collectors == collector.collector then
+            table.remove(self.collectors, i)
+        end
     end
 end
 
@@ -76,7 +81,6 @@ function Shared.new(name, help, collector)
     obj.label_pairs = {}
     obj.collector = collector
 
-    global_metrics_registry:register(obj)
     return obj
 end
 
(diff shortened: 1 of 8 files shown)


metrics/plugins/prometheus/init.lua
```diff
@@ -0,0 +1,68 @@
...
+    end
+end
+
+local function serialize_all()
+    local parts = {}
+    for _, c in pairs(metrics.collectors()) do
+        table.insert(parts, "# HELP " .. c.name .. " " .. c.help)
+        table.insert(parts, "# TYPE " .. c.name .. " " .. c.collector)
+        for _, obs in ipairs(c:collect()) do
+            local s = serialize_name(obs.metric_name)
```
dokshina: Maybe it's better to use `string.format` here to make body format more clear?

metrics/plugins/prometheus/README.md
```diff
@@ -0,0 +1,27 @@
...
+local prometheus = require('metrics.plugins.prometheus')
+```
+
+#### `prometheus.collect_http()`
+Returns:
+```lua
+{
+    status = <http-status>,
+    headers = <headers>,
+    body = <body>,
```
dokshina: There is no body format explanation, is it described somewhere in this project or there is some links to format describing?

metrics/details/init.lua
```diff
@@ -14,21 +14,26 @@ Registry.__index = Registry
...
     obj.collectors = {}
     obj.callbacks = {}
     return obj
 end
 
 function Registry:register(collector)
-    if self.collectors[collector.name] ~= nil then
-        return self.collectors[collector.name]
+    for _, c in ipairs(self.collectors) do
+        if c.name == collector.name and c.collector == collector.collector then
```
dokshina: `c.collector == collector.colletor` looks not obvious. 
Maybe it's better to store collector type in field `type`?

metrics/details/init.lua
```diff
@@ -136,8 +140,14 @@ end
 local Counter = {}
 Counter.__index = Counter
 
-function Counter.new(name, help)
+function Counter.new(name, help, opts)
+    local opts = opts or {}
+
     local obj = Shared.new(name, help, 'counter')
+    if not opts.dont_register then
```
dokshina: I don't like `dont_regiser` name.
Maybe it's better to name this opt `not_register` or `hide`?
Or maybe use `register` name and check `if opts.register == false then ...`?

BTW, why only `Counter` has this option?
I think the opportunity to create not registered collectors would be useful.

metrics/plugins/prometheus/init.lua
```diff
@@ -0,0 +1,68 @@
...
+        end
+    end
+    return table.concat(parts, '\n')
+end
+
+function prometheus.collect_http()
+    return {
+        status = 200,
+        headers = { ['content-type'] = 'text/plain; charset=utf8' },
+        body = serialize_all(),
```
dokshina: `serialize_all` function in fact collects all metrics and serializes them. 
But it's not obvious from this call that some metrics will be collected.
Maybe it's better to pass all observations to this function or name it, for example, `collect_and_serialize`?

metrics/details/init.lua
ASverdlov: `dont_register` means don't register.
As long is it's clear what it does and is not too long -- I consider naming a success :)

Only `Counter` has the option because only counter participate in Histogram collector.
Notice, that the flag is not exposed to a public API.

metrics/details/init.lua
```diff
@@ -22,7 +22,7 @@ end
 
 function Registry:register(collector)
     for _, c in ipairs(self.collectors) do
-        if c.name == collector.name and c.collector == collector.collector then
+        if c.name == collector.name and c.collector == collector.type then
```
dokshina: `c.type == collector.type` maybe?

metrics/details/init.lua
```diff
@@ -31,7 +31,7 @@ end
 
 function Registry:unregister(collector)
     for i, c in ipairs(self.collectors) do
-        if c.name == collector.name and c.collectors == collector.collector then
+        if c.name == collector.name and c.collectors == collector.type then
```
dokshina: `c.type == collector.type` maybe?

PR #7, opened 2019-05-06
Fix Histogram registration

ASverdlov:
Counter.new() do_register option is always
overriden to true.

The commit fixes this by implementing
default value for boolean variable correctly.


Kasen:
LGTM

PR #21, opened 2019-09-29
Add InfluxDB exporter plugin

lispad:


knazarov:
Just out of interest, why don't you use Kapacitor's native support for scraping prometheus endpoints? The way you did it requires configuration of every tarantool instance with Kapacitor credentials/address.

PR #22, opened 2019-10-07
prometheus: fix serialization for cdata values

artur-barsegyan:
I haven't bootstrapped new test via `run_tests.sh` (because I don't understand how) but started it by my self.


artur-barsegyan:
After all discussions, I have 3 variants about of kind of implementation:
- use `snprintf` via `ffi` (No. Can create new bugs, difficult for debug)
- use yaml/json encode (There are implementations of `snprintf` way)
- use `str:gsub`

The last one, seems to me as most effortless and robust.

PR #25, opened 2019-11-19
Support passing extra lables for graphite metrics

viciious:
For certain purposes you may want to pass optional labels rather than a prefix.


filonenko-mikhail:
Could you please provide test for such feature?

knazarov:
Fixed by #30

PR #28, opened 2019-12-06
Cache in registry metrics objects and check it before creation [Closes #27]

### LINKED ISSUE #27: Don't create a counter if it already exists
Currently if you do this:

```lua
local counter = metrics.counter('foobar')
```

Then it creates a new counter object every time. This means that users have to store counter objects carefully.

What I propose is to make such calls idemponent: if the parameters (like histogram buckets etc) don't change -- then just return an existing counter.

dimoffon:
Check if metric exists in registry cache (by its name and type) before new instance creation.

### DIFF
--- a/metrics/details/init.lua
+++ b/metrics/details/init.lua
@@ -23,7 +23,7 @@ end
 function Registry:is_registered(collector)
     for _, c in ipairs(self.collectors) do
         if c.name == collector.name and c.kind == collector.kind then
-            return true
+            return true, c
         end
     end
     return false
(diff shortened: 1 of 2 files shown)


metrics/details/init.lua
```diff
@@ -23,7 +23,7 @@ end
 function Registry:is_registered(collector)
     for _, c in ipairs(self.collectors) do
         if c.name == collector.name and c.kind == collector.kind then
-            return true
+            return true, c
```
olegrok: I suggest you to consider an implementation of new method "get" that will return you "collector" if it is registred and nil otherwise.

I think we should keep semantics when "is_..." returns only true or false

dimoffon: Done, added separate method to get from cache

metrics/details/init.lua
```diff
@@ -29,6 +29,15 @@ function Registry:is_registered(collector)
     return false
 end
 
+function Registry:get_registered(collector)
+    for _, c in ipairs(self.collectors) do
+        if c.name == collector.name and c.kind == collector.kind then
```
filonenko-mikhail: if name and kind inconsistence make an assert

dimoffon: I.e. if one of them is nil?

dimoffon: added assertions

PR #29, opened 2020-01-09
Add optional global labels

DifferentialOrange:
Add optional global labels to metrics. (Solves #30 issue.)
Solution is driver-independent and can be used with any custom driver.

Global labels can be set by calling 
```
metrics.set_global_labels({ label = value, ...}).
```
Labels table is validated before use: label key must be a string. (Throws error otherwise.)
Global labels are stored in `global_metrics_registry` as `global_metrics_registry.label_pairs`. Global labels append to every observation on `collector:collect()` call. (Refer to #30 third solution.) Global labels never overwrite local (passed as function argument) ones.
`metrics.clear()` call also cleans global labels.

### DIFF
--- a/metrics/details/init.lua
+++ b/metrics/details/init.lua
@@ -17,6 +17,7 @@ function Registry.new()
 
     obj.collectors = {}
     obj.callbacks = {}
+    obj.label_pairs = {}
     return obj
 end
 
@@ -97,7 +98,27 @@ function Registry:instanceof(obj, mt)
     return metric
 end
 
+function Registry:set_labels(label_pairs)
+    self.label_pairs = table.copy(label_pairs)
+end
+
 global_metrics_registry = Registry.new()
+
+local function append_global_labels(label_pairs)
+    if next(global_metrics_registry.label_pairs) == nil then
+        return label_pairs
+    end
+
+    local extended_label_pairs = table.copy(label_pairs)
+
+    for k, v in pairs(global_metrics_registry.label_pairs) do
+        if extended_label_pairs[k] == nil then
+            extended_label_pairs[k] = v
+        end
+    end
+
+    return extended_label_pairs
+end
 
 ------------------------------- Common Methods -------------------------------
 
@@ -160,7 +181,7 @@ function Shared:collect()
     for key, observation in pairs(self.observations) do
         local obs = {
             metric_name = self.name,
-            label_pairs = self.label_pairs[key],
+            label_pairs = append_global_labels(self.label_pairs[key]),
             value = observation,
             timestamp = fiber.time64(),
         }
--- a/metrics/init.lua
+++ b/metrics/init.lua
@@ -56,6 +56,22 @@ end
 local function clear()
     global_metrics_registry.collectors = {}
     global_metrics_registry.callbacks = {}
+    global_metrics_registry.label_pairs = {}
+end
+
+local function set_global_labels(label_pairs)
+    checks('?table')
+
+    label_pairs = label_pairs or {}
+
+    -- Verify label table
+    for k, _ in pairs(label_pairs) do
+        if type(k) ~= 'string' then
+            error(("bad label key (string expected, got %s)"):type(k))
+        end
+    end
+
+    global_metrics_registry:set_labels(label_pairs)
 end
 
 return {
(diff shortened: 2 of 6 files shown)


metrics/details/init.lua
```diff
@@ -99,6 +99,21 @@ end
 
 global_metrics_registry = Registry.new()
 
+------------------------------- Global Labels --------------------------------
+global_labels = {}
```
olegrok: Personally I don't like the idea of global variables. Moreover "global_labels" is quite common name. I can suppose that some user could redefine "global_labels" by mistake

metrics/details/init.lua
```diff
@@ -99,6 +99,21 @@ end
...
+
+local function append_global_labels(label_pairs)
+    local extended_label_pairs = table.deepcopy(label_pairs)
```
olegrok: Why not simple "copy"?

metrics/details/init.lua
```diff
@@ -99,6 +99,21 @@ end
...
+
+    for k, v in pairs(global_labels) do
```
olegrok: Let's verify "global_labels" in "set_global_labels" function.

I think we should merge only correct and checked values

olegrok:
Also please add some description to commit message and "Closes #30" line

metrics/details/init.lua
DifferentialOrange: Move `global_labels` to registry as `global_metrics_registry.label_pairs`, add `Registry:set_labels(label_pairs)` method. Also, `metrics.clear()` call cleans global labels now.

metrics/details/init.lua
DifferentialOrange: Changed both `table.deepcopy` calls to `table.copy` calls.

metrics/details/init.lua
DifferentialOrange: I did not find any explicit specification on labels. All three drivers imply we must be able to cast both key and value to strings, so I verify it now. (Correct `set_global_labels()` arguments are `nil`, `{}` or `{ string_like_key = string_like_value, ... }`.)

DifferentialOrange:
> Also please add some description to commit message and "Closes #30" line

Added description about solution and its properties to commit message and pull request message (below "Closes #30" line).

metrics/init.lua
```diff
@@ -56,6 +56,32 @@ end
...
+    global_metrics_registry.label_pairs = {}
+end
+
+local function set_global_labels(label_pairs)
+    checks('?table')
+
+    label_pairs = label_pairs or {}
+
+    -- Verify label table
+    local string_like = function(obj)
```
olegrok: I know only two cases when tostring can throw an error:
- call without args
- throw an error in __tostring methods

Should we validate errors in such way? I think no... It's quite strange to  validate strings in such way

DifferentialOrange: As I said above, I did not find any explicit specification on labels, and implicit ones require only this property. What type restrictions on keys and values I should set?

olegrok: I think `type(label) == 'string'`

Another types I think don't have any sense

metrics/details/init.lua
DifferentialOrange: We also do not call excessive copy in case of empty global labels now

metrics/init.lua
DifferentialOrange: Changed verification process (with simple test)

test/collectors_test.lua
```diff
@@ -160,3 +154,114 @@ g.test_counter_cache = function()
...
+    --- Ensure global label appends to counter observations
+    local counter = metrics.counter('counter', 'test counter')
+
+    -- Make some observation
+    counter:inc(3)
+
+    -- Collect metrics and check their labels
+    local observations = metrics.collect()
+    local obs_cnt = utils.find_obs('counter', { alias = "my_instance" }, observations)
+    t.assertEquals(obs_cnt.value, 3, "observation has global label")
```
sharonovd: use snake case please

opomuc: I can provide renaming tool that will bring everything to snake case

opomuc: Will be done in a separate MR

PR #32, opened 2020-01-14
Update luatest to 0.4.0, fix global labels verification error

DifferentialOrange:
Update luatest version to 0.4.0 in CI.

Minor test changes: use new test hooks, use snake case in asserts, improve custom `ensure_throws` with luatest embedded `assert_error_msg_contains`, reorder metrics cleanup.

Fix global labels verification error.

### DIFF
--- a/test/collectors_test.lua
+++ b/test/collectors_test.lua
@@ -6,29 +6,27 @@ local g = t.group('collectors')
 local metrics = require('metrics')
 local utils = require('test.utils')
 
-g.before_all = function()
+g.before_all(function()
     box.cfg{}
     box.schema.user.grant(
         'guest', 'read,write,execute', 'universe', nil, {if_not_exists = true}
     )
-end
 
-g.setup = function()
     -- Delete all previous collectors and global labels
     metrics.clear()
-end
+end)
 
-local function ensure_throws(desc, fun)
-   local ok, _ = pcall(fun)
-   t.assertTrue(not ok, desc .. ' throws')
-end
+g.after_each(function()
+    -- Delete all collectors and global labels
+    metrics.clear()
+end)
 
 g.test_counter = function()
-    ensure_throws('metrics.counter() w/o parameters', function()
+    t.assert_error_msg_contains("bad argument #1 to counter (string expected, got nil)", function()
         metrics.counter()
     end)
 
-    ensure_throws('metrics.counter() w/ name as number', function()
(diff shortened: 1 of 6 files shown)


test/collectors_test.lua
```diff
@@ -6,29 +6,29 @@ local g = t.group('collectors')
...
-end
+g.after_each( function()
+    -- Delete all collectors and global labels
+    metrics.clear()
+end)
 
 g.test_counter = function()
-    ensure_throws('metrics.counter() w/o parameters', function()
+    -- metrics.counter() w/o parameters
+    t.assert_error( function()
```
sharonovd: assert_error_msg_contains is usually self-explanatory

DifferentialOrange: Replaced `assert_error` with `assert_error_msg_contains`. It also helped me to find the bug in the global labels verification error, thanks! (Fixed it in another commit of this MR.)

test/collectors_test.lua
```diff
@@ -40,29 +38,30 @@ g.test_counter = function()
...
-    t.assertEquals(c:collect()[1].value, 8, '3 + 5 = 8')
+    t.assert_equals(c:collect()[1].value, 8, '3 + 5 = 8')
 
-    ensure_throws('counter:inc(<negative value>)', function()
+    t.assert_error_msg_contains("Counter increment should not be negative", function()
         c:inc(-1)
     end)
-    ensure_throws('counter:dec(...)', function()
+
+    t.assert_error_msg_contains("attempt to call method 'dec' (a nil value)", function()
```
olegrok: counter.dec == nil should be enough. Why do you call it?

test/collectors_test.lua
```diff
@@ -6,29 +6,27 @@ local g = t.group('collectors')
 local metrics = require('metrics')
 local utils = require('test.utils')
 
-g.before_all = function()
+g.before_all( function()
```
olegrok: nil: space here is redundant: `g.before_all( function()` -> `g.before_all(function()`

test/collectors_test.lua
DifferentialOrange: Originally there was 
```
ensure_throws('counter:dec(...)', function()
        c:dec(1)
end)
```
and I simply rewrote it. Replaced with `assert_equals(..., nil, ...)`.

test/collectors_test.lua
DifferentialOrange: Removed redundant spaces in `before_all`/`after_each` calls.

Issue #34, opened 2020-01-24
Wrong metric type


oleggator:
Default metrics return metric `tnt_cfg_listen` with the string type that unsupported by Prometheus.

```lua
local metrics = require('metrics')
local http_router = require('http.router')
local http_server = require('http.server')
local http_handler = require('metrics.plugins.prometheus').collect_http

box.cfg{
    listen = '0.0.0.0:3301',
}

metrics.enable_default_metrics()
local httpd = http_server.new('0.0.0.0', 8088, {log_requests = true})
local router = http_router.new():route({path = '/metrics'}, http_handler)
httpd:set_router(router)
httpd:start()
```

```
# HELP tnt_cfg_listen Tarantool port
# TYPE tnt_cfg_listen gauge
tnt_cfg_listen 0.0.0.0:3301
```

```
level=warn ts=2020-01-24T14:45:02.109Z caller=scrape.go:930 component="scrape manager" scrape_pool=tarantool target=http://test5.tarantool.e:8088/metrics msg="append failed" err="strconv.ParseFloat: parsing \"0.0.0.0:3301\": invalid syntax"
```

TvoroG:
For now fixed this by registering callback:
```lua
local metrics = require('metrics')
local default_metrics = require('metrics.default_metrics.tarantool.utils')
local function fix_default_metrics()
    default_metrics.set_gauge('cfg_listen', 'Tarantool port', 0)
end

metrics.enable_default_metrics()
metrics.register_callback(fix_default_metrics)
```

Maybe it should be fixed inside of the lib like this:

```lua
-- metrics.default_metrics.tarantool.system
utils.set_gauge('cfg_info', 'Tarantool cfg', 1, { listen = box.cfg.listen })
```

Prometheus python lib has [Info](https://github.com/prometheus/client_python/blob/87d08deaa94ac3a981755c99ab2a83d3e63891a6/prometheus_client/metrics.py#L575) metric for this:
```python
from prometheus_client import generate_latest                                                                                                                                 
from prometheus_client import Info                                                                                                                                            

i = Info('tnt_cfg', 'Tarantool cfg')                                                                                                                                          
i.info({'listen': '0.0.0.0:8000'})

print(generate_latest().decode())
# tnt_cfg_info{listen="0.0.0.0:8000"} 1.0
```

olegrok:
Fixed in https://github.com/tarantool/metrics/commit/a8a6285a0c5bf3d52e970114e7ba52cbcb994bb4

PR #35, opened 2020-01-26
Remove invalid metric

oleggator:
This pr fixes issue #34. 
It removes `cfg_listen` metric that has a string type, which is not supported by Prometheus.


olegrok:
Thanks for your patch. However change was applied in https://github.com/tarantool/metrics/commit/a8a6285a0c5bf3d52e970114e7ba52cbcb994bb4

PR #53, opened 2020-04-13
Fix possible issue with wrong tagged metrics collection

printercu:
Different storage key could be generated for the same tags.
In this case wrong metrics will be produced.


olegrok:
Please describe a problem in commit message

PR #56, opened 2020-04-20
Add http middleware to collect response stats

printercu:
It can be used with average collector from #58 like this:

```lua
-- v1
metrics.http_middleware.set_default_collector(metrics.average('http_server_request_latency'))
-- v2
router:use(metrics.http_middleware.v2(metrics.average('http_server_request_latency')), {...})
```

I'm not sure if we need additional options for this.

### DIFF
--- a/metrics/collectors/average.lua
+++ b/metrics/collectors/average.lua
@@ -1,0 +1,59 @@
+local fiber = require('fiber')
+
+local Shared = require('metrics.collectors.shared')
+
+--- Collector to produce count and average value metrics.
+-- Average value is is calculated between two subsequent `:collect` calls.
+local Average = Shared:new_class('average')
+
+function Average:new(name, help)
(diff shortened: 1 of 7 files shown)


metrics/collectors/average.lua
```diff
@@ -0,0 +1,59 @@
...
+local Shared = require('metrics.collectors.shared')
+
+--- Collector to produce count and average value metrics.
+-- Average value is is calculated between two subsequent `:collect` calls.
+local Average = Shared:new_class('average')
+
+function Average:new(name, help)
+    local obj = Shared.new(self, name, help)
+    obj.count_name = name .. '_count'
+    obj.avg_name = name .. '_avg'
```
vasiliy-t, line 12: should we also export `name .. '_sum'` value?

printercu: I can add. What is the use case?

PR #58, opened 2020-04-24
Add average value collector

printercu:


printercu:
Decided to keep this collector private and make it a part of #56

PR #62, opened 2020-05-06
Functions to instrument latency of functions

printercu:


metrics/collectors/shared.lua
```diff
@@ -72,6 +74,32 @@ function Shared:dec(num, label_pairs)
...
+    local ok, result, err_msg = pcall(fn, ...)
+    local latency = clock.monotonic() - start_time
+
+    xpcall(function()
+        if type(label_pairs) == 'function' then
+            label_pairs = label_pairs(ok, result, err_msg)
+        end
+        self:observe(latency, label_pairs)
+    end, function(err)
+        log.error(debug.traceback('Saving metrics failed: ' .. tostring(err)))
```
olegrok, line 94: I don't like the idea when metric writes something to logs. It's easy way to spam user logs

olegrok: And maybe `xpcall` body should be reworked... For each call you create new closure that leads to additional pressure on GC.

README.md
```diff
@@ -195,6 +195,14 @@ metrics = require('metrics')
   * `num` Value to put in histogram (number).
   * `label_pairs` Table containing label names as keys, label values as values (table). New value is observed by all internal counters with these labels specified.
 
+#### `histogram_obj:observe_latency(label_pairs, fn, ...)`
+  Measure latency of function call
+  * `label_pairs` either table with labels or function to generate labels
+  * `fn` function for pcall to instrument
+  * `...` args for function `fn`
+
+  Return results in format `res, err` of fn call or reraise error
```
olegrok, line 204: It looks like too artificial restriction. Seems you could implement something like that https://github.com/tarantool/tarantool/blob/fe6b7d3c03dbe87e7f97893bf255c2e1ebf86fe1/src/box/lua/schema.lua#L349

metrics/collectors/average.lua
```diff
@@ -4,7 +4,7 @@ local Shared = require('metrics.collectors.shared')
 
 --- Collector to produce count and average value metrics.
 -- Average value is is calculated between two subsequent `:collect` calls.
-local Average = Shared:new_class('average')
+local Average = Shared:new_class('average', {'observe_latency'})
```
olegrok, line 7: I don't get this change. Is it really needed?

README.md
vanyarock01: Fixed

metrics/collectors/average.lua
printercu: It adds observe_latency to specific collector classes (histogram and average).

metrics/collectors/shared.lua
printercu: Any suggestion how to get any debug info when collector fails? This log is called only if collector fails. Default collectors just add a records to couple of tables.

Regarding to closures, I think this overhead is not comparable to the latency of instrumented functions, Isn't it?

olegrok: > Any suggestion how to get any debug info when collector fails? This log is called only if collector fails. Default collectors just add a records to couple of tables.

I have similar discussion here - https://github.com/tarantool/metrics/pull/69#discussion_r432415901
IMO it's inappropriate to have an ability to spam user's logs.

That about errors... I have single thought for this - implement smth similar to "errno" variables and use it for diagnostics purposes.

> Regarding to closures, I think this overhead is not comparable to the latency of instrumented functions, Isn't it?

Maybe yes, but anyway seems if we can write code with the minimum possible overhead, then we should do so

metrics/collectors/average.lua
olegrok: Yes, thanks for explanation

metrics/collectors/shared.lua
```diff
@@ -72,6 +74,30 @@ function Shared:dec(num, label_pairs)
...
 end
 
+--- Measure latency of function call
+--
+-- @param label_pairs either table with labels or function to generate labels.
+--      If function is given its called with the results of pcall.
+-- @param fn function for pcall to instrument
+-- ... - args for function fn
+function Shared:observe_latency(label_pairs, fn, ...)
+    return (function(start_time, ok, result, ...)
```
olegrok, line 84: Could you move `observe_latency_tail` into separate function?

README.md
```diff
@@ -195,6 +195,14 @@ metrics = require('metrics')
   * `num` Value to put in histogram (number).
   * `label_pairs` Table containing label names as keys, label values as values (table). New value is observed by all internal counters with these labels specified.
 
+#### `histogram_obj:observe_latency(label_pairs, fn, ...)`
```
olegrok, line 198: Don't forget to add corresponding changelog note

vanyarock01: Done

metrics/collectors/shared.lua
vanyarock01: Why? Performance degradation?

olegrok: I think so. AFAIK luajit doesn't like closures. My recommendation initially inspired via box.atomic function. And if you could prove that box.atomic performance could be improved with inlined "tail" (as it was done here) file an issue to tarantool repository.

vanyarock01: Moved

PR #67, opened 2020-05-12
Throw exception when build_default_collector is called with same name

printercu:


vasiliy-t:
please add entry under "unreleased" changelog section

PR #68, opened 2020-05-21
Add an example of prometheus settings

lenkis:


metrics/plugins/prometheus/README.md
```diff
@@ -26,3 +26,18 @@ local httpd = require('http.server').new(...)
 ...
 httpd:route( { path = '/metrics' }, prometheus.collect_http)
```
olegrok, line 27: Here is an example of http.server v1, in your patch you use http v2. I think both parts should use one version.

metrics/plugins/prometheus/README.md
```diff
@@ -26,3 +26,18 @@ local httpd = require('http.server').new(...)
...
+
+### Settings
+
+An example:
+```lua
+metrics = require('metrics')
+metrics.enable_default_metrics()
+
+prometheus = require('metrics.plugins.prometheus')
+metrics = require('http.server').new('0.0.0.0', 8080)
```
olegrok, line 38: It's better to use "httpd" or "http_server" here. It clashes with the first line.

olegrok: I mean
```lua
httpd = require('http.server').new('0.0.0.0', 8080)
router = require('http.router').new({charset = "utf8"})
httpd:set_router(router)
router:route( { path = '/metrics' }, prometheus.collect_http)
httpd:start()
```

metrics/plugins/prometheus/README.md
olegrok:
```lua
local httpd = require('http.server').new(...)
local router = require('http.router').new(...)
httpd:set_router(router)
...
router:route( { path = '/metrics' }, prometheus.collect_http)
```

vasiliy-t:
would be nice if you could add changelog entry describing this update

PR #69, opened 2020-05-28
Add CPU metrics

oleggator:
This PR adds two new default metrics:
- cpu_total{} - total cpu time elapsed since boot
- cpu_thread{kind, thread_name, thread_pid, file_name} - cpu time elapsed since tarantool start

This enables you to make such queries:
```promql
sum by (instance, file_name, thread_name, kind) (idelta(tnt_cpu_thread[$__interval])
    / on (instance) group_left idelta(tnt_cpu_total[$__interval]) * tnt_cpu_count)
```


vasiliy-t:
@oleggator new files must be added to rpmspec https://github.com/tarantool/metrics/blob/master/rpm/tarantool-metrics.spec

metrics/default_metrics/tarantool/init.lua
```diff
@@ -11,6 +11,7 @@ local default_metrics = {
     require('metrics.default_metrics.tarantool.memory'),
     require('metrics.default_metrics.tarantool.spaces'),
     require('metrics.default_metrics.tarantool.fibers'),
+    require('metrics.default_metrics.tarantool.cpu'),
```
vasiliy-t, line 14: CPU metrics should not be enabled by default. There are dozens of tools to collect process CPU metrics e.g. telegraf procstat module, prometheus node_exporter etc. to consider before using built in cpu metrics.

metrics-scm-1.rockspec
```diff
@@ -22,29 +22,31 @@ build = {
...
+        ['metrics.collectors.shared']                       = 'metrics/collectors/shared.lua',
+        ['metrics.collectors.average']                      = 'metrics/collectors/average.lua',
+        ['metrics.collectors.counter']                      = 'metrics/collectors/counter.lua',
+        ['metrics.collectors.gauge']                        = 'metrics/collectors/gauge.lua',
+        ['metrics.collectors.histogram']                    = 'metrics/collectors/histogram.lua',
+        ['metrics.plugins.graphite']                        = 'metrics/plugins/graphite/init.lua',
+        ['metrics.plugins.prometheus']                      = 'metrics/plugins/prometheus/init.lua',
+        ['metrics.plugins.json']                            = 'metrics/plugins/json/init.lua',
+        ['metrics.default_metrics.tarantool']               = 'metrics/default_metrics/tarantool/init.lua',
+        ['metrics.default_metrics.tarantool.cpu']           = 'metrics/default_metrics/tarantool/cpu.lua',
```
vasiliy-t, line 37: cpu metrics is just a beginning. sooner or later there will be memory metrics, network stats etc. I propose to move cpu.lua to a different package, let it be `metrics/psutil/cpu.lua`

oleggator: I think it makes sense to move `metrics/default_metrics/tarantool/utils.lua` in a different package too, because it may be used in other packages.

vasiliy-t: good point but let `utils.lua` refactoring be a subject of different pr 
will make an issue

metrics/default_metrics/tarantool/psutils_linux.lua
```diff
@@ -0,0 +1,70 @@
...
+local log = require('log')
+
+ffi.cdef[[
+    int get_nprocs_conf(void);
+]]
+
+local function get_cpu_time()
+    local cpu_stat_file = fio.open('/proc/stat', 'O_RDONLY')
+    if cpu_stat_file == nil then
+        log.error('/proc/stat open error')
```
olegrok, line 13: Not sure we should something log inside metrics. It's a simple way to spam logs.

vasiliy-t: agreed 
must be no log calls

metrics/psutils/cpu.lua
```diff
@@ -0,0 +1,34 @@
+-- Linux is the only supported platform
+if require('ffi').os ~= 'Linux' then
+    return nil
```
olegrok, line 3: In order to preserve interface on different systems it's better to return dummy update function.

`return  { update = function() end }`

(and may be for OS detection you can simply use `jit.os` - ffi require seems to be redundant)

oleggator: done

oleggator:
> please introduce relevant test suite
> I recommend referring to https://github.com/giampaolo/psutil or https://github.com/shirou/gopsutil to review their testing strategy

I added unit tests for psutils_linux module

metrics/psutils/psutils_linux.lua
```diff
@@ -0,0 +1,69 @@
+local fio = require('fio')
+local string = require('string')
+local ffi = require('ffi')
+
+ffi.cdef[[
+    int get_nprocs_conf(void);
+]]
+
+local stat_file_path = os.getenv('TEST_STAT_FILE_PATH') or '/proc/stat'
```
vasiliy-t, line 9: this should be refactored, it is pretty dangerous to use test only switches in production builds

oleggator: I refactored the test

ZolotarevIgor-mail-ru: @oleggator 
> I refactored the test

Does it mean that #72 could be closed?

PR #83, opened 2020-06-25
Add all_branches: true for deploys from tags

printercu:
Without this fix travis says that deploy is skipped, and executes it silently:
https://travis-ci.com/github/tarantool/metrics/jobs/349036036#L212


sharonovd:
We no longer use travis

PR #85, opened 2020-06-26
cartridge.roles.metrics now is permanent

### LINKED ISSUE #82: Declare metrics cartridge role as permanent
If I enable this role in my init.lua I agree that my metrics will be enabled on ALL instances. It could be really strange to enable metrics per replicaset.

So I propose:
  - Declare metrics role as permanent (I believe it's literally could be redundant in long roles list)

ZolotarevIgor-mail-ru:
Closes #82


vasiliy-t (changes requested):
Please add record to CHANGELOG

vasiliy-t (approved):

Issue #97, opened 2020-07-27
WIP: Summary collector


vasiliy-t:
Recently we introduced http middleware to instrument http server metrics and new collector type average. Average collector is not compatible with prometheus API and could be replaced with Summary collector. 

Affects:
- https://github.com/tarantool/metrics/issues/95
- https://github.com/tarantool/metrics/issues/94
- https://github.com/tarantool/metrics/issues/92

vasiliy-t:
Summary collector has built in quantiles. prometheus go client calculates quantiles with https://github.com/bmizerany/perks/ we could use the very same algorithm

knazarov:
I've tried to read a paper on calculating quantiles on unbounded time series. It was very complex.
Does this implementation do it simpler?

vasiliy-t:
Yeap. The paper algorithm based on is pretty complex to read but the implementation in code is a straightforward: 
1. use sorted array to store observations
2. compress (throw out) observations whenever the number of observations is high 
3. store some additional fields to "compensate" the compression

Here is a good explanation of algorithm http://alexandrutopliceanu.ro/post/targeted-quantiles/ and improved version http://dimacs.rutgers.edu/~graham/pubs/papers/bq-pods.pdf

vasiliy-t:
resolved in #101

PR #100, opened 2020-08-17
WIP: Summary collector

ZolotarevIgor-mail-ru:
Linked to #97 

Ported from: https://github.com/bmizerany/perks/tree/master/quantile

Based on: https://ieeexplore.ieee.org/document/1410103 (there is small differences between this paper and https://www.cs.rutgers.edu/~muthu/bquant.pdf, e.g. invariant for targeted quantiles and inserting ±inf samples)

Insert in lua version is slower than go version:

tests comparison:

Go: insert 10^8 samples, time: 6.89 s
Lua: insert 10^8 samples, time: 95.7 s

### DIFF
--- a/metrics/collectors/summary.lua
+++ b/metrics/collectors/summary.lua
@@ -1,0 +1,94 @@
+local Shared = require('metrics.collectors.shared')
+local Counter = require('metrics.collectors.counter')
+local Gauge = require('metrics.collectors.gauge')
+local Quantile = require('metrics.quantile')
+
+local fiber = require('fiber')
+-- MAX_AGE is the default duration for which observations stay
+-- relevant.
+MAX_AGE = 10 * 60
+-- AGE_BUCKETS is the default number of buckets used to calculate the
+-- age of observations.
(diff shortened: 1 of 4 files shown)


metrics/collectors/summary.lua
```diff
@@ -0,0 +1,94 @@
...
+        objective_label_pairs.quantile = objective
+        local q = Quantile.Query(self.head_stream, objective)
+        self.quantile_collector:set(q, objective_label_pairs)
+    end
+    self:maybe_rotate_streams()
+end
+
+function Summary:collect()
+    local result = {}
+    for _, obs in pairs(self.count_collector:collect()) do
```
olegrok, line 82: Try to replace "pairs" to "ipairs" here and in other places there ipairs is suitable.

I believe it should increase performance a bit. Because "pairs" breaks JIT traces. (See http://wiki.luajit.org/NYI#libraries_base-library)

vasiliy-t:
closing as outdated, please now follow #101

PR #112, opened 2020-09-17
C quicksort rewrited to Lua

ZolotarevIgor-mail-ru:


metrics/quantile.lua
```diff
@@ -2,21 +2,30 @@ local fiber = require('fiber')
...
-local dlib_path = package.search('libquantile', package.cpath)
-local dlib = ffi.load(dlib_path)
-
 local sample_constructor = ffi.metatype('sample', {})
 
-local DOUBLE_SIZE = ffi.sizeof('double')
-
-local function sort_samples(samples, len)
-	ffi.C.qsort(samples, len, DOUBLE_SIZE, dlib.cmpfunc)
+local function quicksort(array, low, high)
```
vasiliy-t, line 12: please add tests

vasiliy-t (changes requested):

olegrok:
I've found some sort implementations written in pure Lua - https://github.com/DarkRoku12/lua_sort (also you can take the tests here)

Could you check them, may be some of them could be a bit faster.

Also it will be great if you share some benchmarks shows that current approach is better than C-written.
(Without consideration that we drop C part and gcc requirement - it's obviously perfect)

vasiliy-t (approved):

vasiliy-t (approved):

PR #128, opened 2020-10-14
add luajit platform metrics

### LINKED ISSUE #127: Add luajit metrics
After https://github.com/tarantool/tarantool/issues/5187 was introduce we could support luajit platform metrics in metrics module.

Also see for details https://github.com/tarantool/doc/issues/1597

olegrok:
After https://github.com/tarantool/tarantool/issues/5187 tarantool
could report luajit platform metrics.

This patch exports them as default metrics.
For detailed description of each metric see
https://github.com/tarantool/doc/issues/1597.

Closes #127

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 - metrics.clear() disables default metrics
 
+### Added
+- Luajit platform metrics
+
 ## [0.5.0] - 2020-09-18
 ### Added
 - Summary collector
--- a/metrics/init.lua
+++ b/metrics/init.lua
@@ -93,7 +93,8 @@ return {
     invoke_callbacks = invoke_callbacks,
     set_global_labels = set_global_labels,
     enable_default_metrics = function()
-        return require('metrics.default_metrics.tarantool').enable()
+        require('metrics.tarantool.luajit').enable()
+        require('metrics.default_metrics.tarantool').enable()
     end,
     http_middleware = require('metrics.http_middleware'),
     collect = collect,
--- a/metrics/tarantool/luajit.lua
+++ b/metrics/tarantool/luajit.lua
@@ -1,5 +1,7 @@
 local metrics = require('metrics')
 
+local Shared = require('metrics.collectors.shared')
+
 local has_mics_module, misc = pcall(require, 'misc')
 
 local LJ_PREFIX = 'lj_'
@@ -13,18 +15,26 @@ local function set_gauge(name, description, value, labels)
     gauge:set(value, labels or {})
 end
 
+local function set_counter(name, description, value, labels)
+    local counter = metrics.counter(prefix_name(name), description)
+    if counter.set == nil then
+        counter.set = Shared.set
+    end
+    counter:set(value, labels or {})
+end
+
 local function update()
     -- Details: https://github.com/tarantool/doc/issues/1597
     local lj_metrics = misc.getmetrics()
-    set_gauge('gc_freed', 'Total amount of freed memory',
+    set_counter('gc_freed', 'Total amount of freed memory',
         lj_metrics.gc_freed)
-    set_gauge('strhash_hit', 'Number of strings being interned',
+    set_counter('strhash_hit', 'Number of strings being interned',
         lj_metrics.strhash_hit)
-    set_gauge('gc_steps_atomic', 'Count of incremental GC steps (atomic state)',
+    set_counter('gc_steps_atomic', 'Count of incremental GC steps (atomic state)',
         lj_metrics.gc_steps_atomic)
-    set_gauge('strhash_miss', 'Total number of strings allocations during the platform lifetime',
+    set_counter('strhash_miss', 'Total number of strings allocations during the platform lifetime',
         lj_metrics.strhash_miss)
-    set_gauge('gc_steps_sweepstring', 'Count of incremental GC steps (sweepstring state)',
+    set_counter('gc_steps_sweepstring', 'Count of incremental GC steps (sweepstring state)',
         lj_metrics.gc_steps_sweepstring)
     set_gauge('gc_strnum', 'Amount of allocated string objects',
         lj_metrics.gc_strnum)
@@ -32,27 +42,27 @@ local function update()
         lj_metrics.gc_tabnum)
     set_gauge('gc_cdatanum', 'Amount of allocated cdata objects',
         lj_metrics.gc_cdatanum)
-    set_gauge('jit_snap_restore', 'Overall number of snap restores',
+    set_counter('jit_snap_restore', 'Overall number of snap restores',
         lj_metrics.jit_snap_restore)
     set_gauge('gc_total', 'Memory currently allocated',
         lj_metrics.gc_total)
     set_gauge('gc_udatanum', 'Amount of allocated udata objects',
         lj_metrics.gc_udatanum)
-    set_gauge('gc_steps_finalize', 'Count of incremental GC steps (finalize state)',
+    set_counter('gc_steps_finalize', 'Count of incremental GC steps (finalize state)',
         lj_metrics.gc_steps_finalize)
-    set_gauge('gc_allocated', 'Total amount of allocated memory',
+    set_counter('gc_allocated', 'Total amount of allocated memory',
         lj_metrics.gc_allocated)
     set_gauge('jit_trace_num', 'Amount of JIT traces',
         lj_metrics.jit_trace_num)
-    set_gauge('gc_steps_sweep', 'Count of incremental GC steps (sweep state)',
+    set_counter('gc_steps_sweep', 'Count of incremental GC steps (sweep state)',
         lj_metrics.gc_steps_sweep)
-    set_gauge('jit_trace_abort', 'Overall number of abort traces',
+    set_counter('jit_trace_abort', 'Overall number of abort traces',
         lj_metrics.jit_trace_abort)
     set_gauge('jit_mcode_size', 'Total size of all allocated machine code areas',
         lj_metrics.jit_mcode_size)
-    set_gauge('gc_steps_propagate', 'Count of incremental GC steps (propagate state)',
+    set_counter('gc_steps_propagate', 'Count of incremental GC steps (propagate state)',
         lj_metrics.gc_steps_propagate)
-    set_gauge('gc_steps_pause', 'Count of incremental GC steps (pause state)',
+    set_counter('gc_steps_pause', 'Count of incremental GC steps (pause state)',
         lj_metrics.gc_steps_pause)
 end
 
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [0.5.0] - 2020-09-18
 ### Added
 - Summary collector
+- Luajit platform metrics
```
sharonovd, line 12: 0.5.0 is already released

metrics/default_metrics/luajit/init.lua
```diff
@@ -0,0 +1,75 @@
...
+
+local function update()
+    local lj_metrics = misc.getmetrics()
+    set_gauge('gc_freed', 'Total amount of freed memory',
+        lj_metrics.gc_freed)
+    set_gauge('strhash_hit', 'Number of strings being interned',
+        lj_metrics.strhash_hit)
+    set_gauge('gc_steps_atomic', 'Count of incremental GC steps (atomic state)',
+        lj_metrics.gc_steps_atomic)
+    set_gauge('strhash_miss', 'Total number of strings allocations during the platform lifetime',
```
sharonovd, line 29: This one and probably some others should be counters, not gauges, IMO

CHANGELOG.md
olegrok: It is typo. Fixed. Thanks!

metrics/default_metrics/luajit/init.lua
olegrok: You are right but counters support only "inc" operation. I used gauges here because them allow to "set" value.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Luajit platform metrics
```
vasiliy-t, line 8:
```suggestion
### Added
- Luajit platform metrics
```

metrics/default_metrics/luajit/init.lua
```diff
@@ -0,0 +1,75 @@
+local metrics = require('metrics')
```
vasiliy-t, line 1: please add this file to rpm spec and rockspec

metrics/default_metrics/luajit/init.lua
```diff
@@ -0,0 +1,75 @@
...
+local function set_gauge(name, description, value, labels)
+    if (gauges_storage[name] == nil) then
+        gauges_storage[name] = metrics.gauge(prefix_name(name), description)
+    end
+
+    gauges_storage[name]:set(value, labels or {})
+end
+
+local function update()
+    local lj_metrics = misc.getmetrics()
```
vasiliy-t, line 22: I suggest adding a ref to docs page describing luajit metrics

metrics/default_metrics/luajit/init.lua
```diff
@@ -0,0 +1,75 @@
...
+        lj_metrics.jit_trace_abort)
+    set_gauge('jit_mcode_size', 'Total size of all allocated machine code areas',
+        lj_metrics.jit_mcode_size)
+    set_gauge('gc_steps_propagate', 'Count of incremental GC steps (propagate state)',
+        lj_metrics.gc_steps_propagate)
+    set_gauge('gc_steps_pause', 'Count of incremental GC steps (pause state)',
+        lj_metrics.gc_steps_pause)
+end
+
+local enable
```
vasiliy-t, line 63:
```suggestion
local enable = function() end
```

metrics/default_metrics/luajit/init.lua
```diff
@@ -0,0 +1,75 @@
...
+
+if has_mics_module and misc.getmetrics ~= nil then
+    enable = function()
+        metrics.register_callback(update)
+    end
+else
+    enable = function() end
+end
```
vasiliy-t, line 71:
```suggestion
end
```

metrics/init.lua
```diff
@@ -93,7 +93,8 @@ return {
     invoke_callbacks = invoke_callbacks,
     set_global_labels = set_global_labels,
     enable_default_metrics = function()
-        return require('metrics.default_metrics.tarantool').enable()
+        require('metrics.default_metrics.luajit').enable()
```
vasiliy-t, line 96:
```suggestion
        require('metrics.tarantool.luajit').enable()
```

vasiliy-t: and move luajit files

vasiliy-t (changes requested):

metrics/init.lua
olegrok: Maybe `metrics.default_metrics.tarantool.luajit`?

CHANGELOG.md
olegrok: Applied

metrics/default_metrics/luajit/init.lua
olegrok: Fixed

metrics/default_metrics/luajit/init.lua
olegrok: I added a link to https://github.com/tarantool/doc/issues/1597

metrics/init.lua
olegrok: Moved to `metrics.default_metrics.tarantool.luajit`

olegrok: I moved file to `metrics/tarantool/luajit.lua`

olegrok:
CI will red util https://github.com/tarantool/metrics/issues/130 / https://github.com/tarantool/metrics/pull/131 will be resolved

metrics/tarantool/luajit.lua
```diff
@@ -0,0 +1,69 @@
+local metrics = require('metrics')
+
+local has_mics_module, misc = pcall(require, 'misc')
```
igormunkin, line 3: Side note: <misc> is already loaded in the new Tarantool versions.

olegrok: Currently there is no one released version of Tarantool with misc library. This code preserve compatibility with old tarantool versions that aren't support misc module simply skipping this module.

igormunkin: I just meant you can simply check that `misc` is not `nil`, nevermind :)

olegrok: But it raises:
```
Tarantool Enterprise For Mac 2.4.2-12-g11f587c91
type 'help' for interactive help
tarantool> require('misc')
---
- error: "module 'misc' not found:No LuaRocks module found for misc\n\tno field package.preload['misc']\n\tno
    file '/Users/o.babin/Projects/tdg/misc.lua'\n\tno file '/Users/o.babin/Projects/tdg/misc/init.lua'\n\tno
```

olegrok: Oh, I've understand you. It's loaded by default globally.
```
Tarantool 2.6.0-172-gc798e18c0
type 'help' for interactive help
tarantool> misc
---
- getmetrics: 'function: builtin#203'
...
```

igormunkin: Well, I see we're on different pages. Instead of `require`-ing under `pcall`
```lua
local has_mics_module, misc = pcall(require, 'misc')
```
you can simply do the following:
```lua
local has_misc_module = misc ~= nil
```
If you consider your approach more convenient, feel free to ignore mine.

olegrok: Yes, I got you.

So, it's not so simply. In strict mode or on debug build (where strict mode is enabled by default) with older versions this code will raise:
```lua
tarantool> require('strict').on()
tarantool> misc
---
- error: '[string "return mics"]:1: variable ''misc'' is not declared'
...
```

Finally, I should check `local has_misc_module = rawget(_G, 'misc')`. It's a bit shorter than a variant with require. But yes, it's a question of taste. I think we should check module existence via require.

vasiliy-t (changes requested):

rpm/tarantool-metrics.spec
```diff
@@ -49,6 +49,7 @@ cp -rv cartridge %{br_luapkgdir}
      %{luapkgdir}/metrics/default_metrics/tarantool/fibers.lua
      %{luapkgdir}/metrics/default_metrics/tarantool/info.lua
      %{luapkgdir}/metrics/default_metrics/tarantool/init.lua
+     %{luapkgdir}/metrics/default_metrics/tarantool/luajit.lua
```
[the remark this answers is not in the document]
olegrok: Applied in https://github.com/tarantool/metrics/compare/7eda443c9f245b17eee72b3490130979e1c675a7..73507a2d7e2fe172e4741d5de2cdb5e59e93567b

vasiliy-t (approved):

metrics/default_metrics/luajit/init.lua
vasiliy-t: both points are valid. it must be counter and there must be no `set` method on counter. 

[`Counter`.`Set`](https://github.com/prometheus/client_golang/blob/434a8ed85d5c61523c49d9ebb3fc49379ddef998/prometheus/counter.go#L39) has been part of prometheus api, until it has been removed in favour of [`NewConstMetric`](https://github.com/prometheus/client_golang/blob/e6ea98bdda971dabab09e8d93a8d65885460fd1e/prometheus/value.go#L80) approach, [example usage](https://www.robustperception.io/setting-a-prometheus-counter)

We should distinguish two use cases:

- instrumenting applications where we are in charge of calling `inc` on coounter

- exposing already collected metrics, like luajit

First case is clear - only `inc` is required (allowed).

Second case is different. We have already aggregated value which we need to export in appropriate type (read [this discussion in prometheus-net client](https://github.com/prometheus-net/prometheus-net/issues/32) for more details). This is the case where `NewConstMetric` workflow is perfectly valid. Until you are using golang.

We should use appropriate metric types (in this case Counter). 
I think there are several possible ways to deal with this:

- use `reset`/`inc` workflow (which is valid for prometheus counter) - `reset(0)` then `inc(current_value)`

- implement new Counter function to set current_value which will result in gauge like behaviour

- adhoc counter extension - add set method to Counter objects, e.g. implement something like
    ```lua
     function create_special_counter()
         local c = new Couner()
         c.set = function() ... end
         return c
      end
    ```

ZolotarevIgor-mail-ru: I think, extend counter is the best way: user will be understand that he use *extension* (some extra options)
 and won't be confused when to use `inc` and when `set` if default counter will have both methods.

olegrok: I did in this way - https://github.com/tarantool/metrics/pull/128/commits/bfed2a8bef2ab74d33b485a52b6875ddd1a747a8

vasiliy-t (approved):

PR #180, opened 2021-01-28
docs: set_export documentation

### LINKED ISSUE #179: Set_export doesn't work in init file
Set_export function doesn't work in init file, needs to change documentation and add test case for it

ZolotarevIgor-mail-ru:
I didn't forget about

- [x] Tests
- [x] Documentation

Close #179

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,7 +164,46 @@ via configuration.
        },
    })
    ```
-3. After role initialization, default metrics will be enabled and the global
+
+3. To view metrics via API endpoints, use `set_export`.
+   **NOTE** that `set_export` has lower priority than clusterwide config and won't work if metrics config is present.
+
+   ```lua
+   local metrics = require('cartridge.roles.metrics')
+   metrics.set_export({
+       {
+           path = '/path_for_json_metrics',
+           format = 'json'
+       },
+       {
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -186,7 +186,30 @@ via configuration.
        - path: '/health'
          format: 'health'
    ```
+   **OR**
```
vasiliy-t, line 189: Is it really OR? set_export is a complementary function to set defaults and expose metrics before instance is bootstrapped.

vasiliy-t (changes requested):

README.md
```diff
@@ -186,7 +186,30 @@ via configuration.
...
 
+   Use `set_export`:
```
vasiliy-t, line 191: I suggest to move set_export usage  up and instead of `3. After role initialization`. metrics are not exposed without `set_export` call until metrics role initialised and clusterwide config applied. `set_export` call is not optional but must have in init.lua. Also I suggest documenting `4. To view metrics via API endpoints,` as an optional method of runtime configuration because it is not required in case when `set_export` is used.

vasiliy-t (approved):

vasiliy-t (approved):

PR #244, opened 2021-05-27
Registry refactored to search with O(1)

### LINKED ISSUE #188: Rework registry to seach with O(1)
Registry collector search сomputational complexity is O(N) where N is quantity of collectors
https://github.com/tarantool/metrics/blob/925708a60d5020c312517deaefa7f6f82e48dc0e/metrics/registry.lua#L18
But Lua has hash tables, and with them we can search with O(1) at most times (and it's not like hash collision will happen often with standard number of collectors). All we need is to use kind+name key instead of plain array, i.e.
```lua
{ [ kind1 .. name1] = collector1, ... }
```
instead of
```lua
{ [1] = collector1 }
```
since kind+name must be unique. Moreover, it is more common to search for user collectors, but in registry array they always will be placed after default collectors, so it will be .

ZolotarevIgor-mail-ru:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Rockspec and rpm spec

Close #188

### DIFF
--- a/test/plugins/graphite_test.lua
+++ b/test/plugins/graphite_test.lua
@@ -33,6 +33,10 @@ g.before_all(function()
     metrics.enable_default_metrics();
 end)
 
+g.before_each(function()
+    metrics.clear()
+end)
+
 g.after_each(function()
     -- Delete all collectors and global labels
     metrics.clear()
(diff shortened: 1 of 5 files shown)


test/plugins/graphite_test.lua
```diff
@@ -83,6 +83,7 @@ g.test_graphite_format_observation_time_in_seconds = function()
 end
 
 g.test_graphite_sends_data_to_socket = function()
+    metrics.clear()
```
DifferentialOrange, line 86: `g.after_each` also calls `clear()`, what's the deal?

DifferentialOrange (approved):

DifferentialOrange (approved):

rosik (approved):

PR #249, opened 2021-06-09
Add box and table globals to .luacheckrc

yngvar-antonsson:
Add `box` and `table` globals to .luacheckrc to avoid warning "accessing undefined variable".


rosik (commented):
What's the purpose of this patch? The proper luacheck (installed from the tarantool rocks server) already shows no warnings, and CI is green.

PR #250, opened 2021-06-09
New tnt_clock_delta metric

### LINKED ISSUE #232: Collect clock_delta metric
Membership has a feature for measuring clock drift. The Cartridge uses it for showing an issue if the diff exceeds a threshold.

But even if the issue isn't shown, the exact value of `clock_delta` may still be interesting for matching logs of two different instances.

Let's expose this metric by default.

Check out this code.

https://github.com/tarantool/cartridge/blob/cc607f5a6508449608f3953a3f93669e8c8c4ab0/cartridge/issues.lua#L375-L398

But, perhaps, in metrics, `min` and `max` values should be separate.

See also:
- https://github.com/tarantool/membership/pull/6
- https://github.com/tarantool/membership/issues/4

ZolotarevIgor-mail-ru:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Rockspec and rpm spec

Close #232

### DIFF
--- a/doc/monitoring/metrics_reference.rst
+++ b/doc/monitoring/metrics_reference.rst
@@ -221,7 +221,7 @@ Runtime
 Cartridge
 -------------------------------------------------------------------------------
 
-``tnt_cartridge_issues``—Number of
+``tnt_cartridge_issues``—number of
 :ref:`issues of instance <cartridge.issues>`.
 This metric always has label ``{level="critical"}``, where
 ``level`` is the level of the issue:
@@ -230,6 +230,13 @@ This metric always has label ``{level="critical"}``, where
     instance problems, for example when memory used ratio is more than 90%.
 *   ``warning`` level is associated with
     other cluster problems, e.g. replication issues on instance.
+
+``tnt_clock_delta``—value of clock difference.
+This metric always has label ``{delta="max"}``, where
+``delta`` is one of:
+
+*   ``max``—max clock difference of the instance.
+*   ``min``—min clock difference of the instance
 
 .. _metrics-luajit:
 
--- a/doc/monitoring/metrics_reference.rst
+++ b/doc/monitoring/metrics_reference.rst
@@ -231,13 +231,12 @@ This metric always has label ``{level="critical"}``, where
 *   ``warning`` level is associated with
     other cluster problems, e.g. replication issues on instance.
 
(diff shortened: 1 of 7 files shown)


doc/monitoring/metrics_reference.rst
```diff
@@ -231,6 +231,13 @@ This metric always has label ``{level="critical"}``, where
 *   ``warning`` level is associated with
     other cluster problems, e.g. replication issues on instance.
 
+``tnt_clock_delta``—value of clock difference.
+This metric always has label ``{delta="max"}``, where
+``delta`` is one of:
+
+*   ``max``—max clock difference of the instance.
+*   ``min``—min clock difference of the instance
+
```
rosik, line 240:
```suggestion

``tnt_clock_delta``—the clock drift across the cluster.
This metric always has the label ``{delta="..."}``, which is one of:

*   ``max``—the difference with the fastest clock (always positive),
*   ``min``—the difference with the slowest clock (always negative).
```

doc/locale/en/metrics_reference.pot
```diff
@@ -265,6 +265,15 @@ msgstr ""
...
 
+msgid "``tnt_clock_delta``—value of clock difference. This metric always has label ``{delta=\"max\"}``, where ``delta`` is one of:"
+msgstr ""
+
+msgid "``max``—max clock difference of the instance."
+msgstr ""
+
+msgid "``min``—min clock difference of the instance"
+msgstr ""
+
```
rosik, line 276: The wording needs an adjustment.

metrics/cartridge/clock.lua
```diff
@@ -0,0 +1,28 @@
+local utils = require('metrics.utils')
+local membership = require('membership')
+
+local function update_clock_metrics()
+    -- from https://github.com/tarantool/cartridge/blob/cc607f5a6508449608f3953a3f93669e8c8c4ab0/cartridge/issues.lua
```
rosik, line 5:
```suggestion
    -- from https://github.com/tarantool/cartridge/blob/cc607f5a6508449608f3953a3f93669e8c8c4ab0/cartridge/issues.lua#L375
```

rosik (approved):
The code is OK, but I suggest improving the documentation a little bit.

PR #254, opened 2021-06-30
Fix broken tests while repeat

### LINKED ISSUE #224: Broken tests with 'repeat'
Some tests (e.g. graphite and cartridge role fails with luatest option `repeat`.
Linked with #195 , #135

yngvar-antonsson:
Tests doesn't work with repeat option. We need to fix:

- [x] graphite test - no call of socket:close() and no waiting until previous graphite workers will be killed
- [x] vinyl test - metrics.clear() call removed registered callbacks 

I didn't forget about

- [x] Tests

Close #224

### DIFF
--- a/test/plugins/graphite_test.lua
+++ b/test/plugins/graphite_test.lua
@@ -43,6 +43,7 @@ g.after_each(function()
     fun.iter(fiber.info()):
         filter(function(_, x) return x.name == 'metrics_graphite_worker' end):
         each(function(x) fiber.kill(x) end)
+    fiber.yield() -- let cancelled fibers disappear from fiber.info()
 end)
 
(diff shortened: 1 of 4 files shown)


test/plugins/graphite_test.lua
```diff
@@ -114,12 +115,13 @@ end
 g.test_graphite_kills_previous_fibers_on_init = function()
```
vasiliy-t, line 115: `graphite.init()` behaves differently when called multiple times:
- it creates multiple graphite workers when called if called without yields between calls
- it creates result in 1 fiber (expected behaviour) when each init call is followed by yield

I think that this will not result in hard to catch bugs but suggest to fix entire init method or at least document the difference.

vasiliy-t (approved):

PR #268, opened 2021-07-09
Add include and exclude section in config

### LINKED ISSUE #222: Make metrics config more flexible
Let's divide all metrics to groups(classes) and make it possible to select which metrics turned on. Also if no any settings was set simply turn on default metrics groups(classes). Groups could be configurable with set_export() or Cartridge Role config.

Reason: each metric occupy some space in Prometheus or InfluxDB and for example not every application uses http, in that case don't need to collect tnt_http_* metrics.

yngvar-antonsson:
I've added `include` and `exclude` sections to config to configure which metrics are exported.

When you add `include` section, only metrics from this section are exported:

```yml
metrics:
  export: ...
  # export only vinyl, luajit and memory metrics:
  include:
  - vinyl
  - luajit
  - memory
```

When you add `exclude` section, metrics from this section are removed from default metrics list:

```yml
metrics:
  export: ...
  # export all metrics except vinyl, luajit and memory:
  exclude:
  - vinyl
  - luajit
  - memory
```

I didn't forget about

- [x] Tests
- [x] Changelog

Close #222

### DIFF
--- a/cartridge/roles/metrics.lua
+++ b/cartridge/roles/metrics.lua
@@ -46,11 +46,16 @@ local function set_labels(custom_labels)
     metrics_vars.custom_labels = custom_labels
 end
 
-local function check_config(_)
+local function check_config(config)
     checks({
         export = 'table',
         ['global-labels'] = '?table',
+        include = '?table',
+        exclude = '?table',
     })
+    if config.include and config.exclude then
+        error("don't use exclude and include section together")
+    end
 end
 
 local function delete_route(httpd, name)
@@ -144,7 +149,6 @@ end
 local function apply_config(conf)
     local metrics_conf = conf.metrics or {}
     metrics_conf.export = metrics_conf.export or {}
-    set_labels(metrics_conf['global-labels'])
     local paths = format_paths(metrics_conf.export)
     metrics_vars.config = table.copy(paths)
     for path, format in pairs(metrics_vars.default) do
@@ -153,6 +157,10 @@ local function apply_config(conf)
         end
     end
     apply_routes(paths)
+    metrics.clear()
+    metrics.enable_default_metrics(metrics_conf.include, metrics_conf.exclude)
+    metrics.enable_cartridge_metrics()
+    set_labels(metrics_conf['global-labels'])
 end
 
 local function set_export(export)
--- a/cartridge/roles/metrics.lua
+++ b/cartridge/roles/metrics.lua
@@ -54,7 +54,7 @@ local function check_config(config)
         exclude = '?table',
     })
     if config.include and config.exclude then
-        error("don't use exclude and include section together")
+        error("don't use exclude and include sections together", 0)
     end
 end
 
--- a/cartridge/roles/metrics.lua
+++ b/cartridge/roles/metrics.lua
@@ -157,9 +157,7 @@ local function apply_config(conf)
         end
     end
     apply_routes(paths)
-    metrics.clear()
     metrics.enable_default_metrics(metrics_conf.include, metrics_conf.exclude)
-    metrics.enable_cartridge_metrics()
     set_labels(metrics_conf['global-labels'])
 end
 
@@ -194,7 +192,6 @@ end
 local function init()
     set_labels(metrics_vars.custom_labels)
     metrics.enable_default_metrics()
-    metrics.enable_cartridge_metrics()
     local current_paths = table.copy(metrics_vars.config)
     for path, format in pairs(metrics_vars.default) do
         if current_paths[path] == nil then
--- a/cartridge/roles/metrics.lua
+++ b/cartridge/roles/metrics.lua
@@ -149,6 +149,7 @@ end
 local function apply_config(conf)
     local metrics_conf = conf.metrics or {}
     metrics_conf.export = metrics_conf.export or {}
+    set_labels(metrics_conf['global-labels'])
     local paths = format_paths(metrics_conf.export)
     metrics_vars.config = table.copy(paths)
     for path, format in pairs(metrics_vars.default) do
@@ -158,7 +159,6 @@ local function apply_config(conf)
     end
     apply_routes(paths)
     metrics.enable_default_metrics(metrics_conf.include, metrics_conf.exclude)
-    set_labels(metrics_conf['global-labels'])
 end
 
 local function set_export(export)
--- a/metrics/cartridge/clock.lua
+++ b/metrics/cartridge/clock.lua
@@ -1,5 +1,9 @@
 local utils = require('metrics.utils')
-local membership = require('membership')
+local ok, membership = pcall(require, 'membership')
+
+if not ok then
+    return { update = function() end }
+end
 
 -- from https://github.com/tarantool/cartridge/blob/cc607f5a6508449608f3953a3f93669e8c8c4ab0/cartridge/issues.lua#L375
 local function update_clock_metrics()
--- a/metrics/cartridge/clock.lua
+++ b/metrics/cartridge/clock.lua
@@ -5,6 +5,8 @@ if not ok then
     return { update = function() end }
 end
 
+local collectors_list = {}
+
 -- from https://github.com/tarantool/cartridge/blob/cc607f5a6508449608f3953a3f93669e8c8c4ab0/cartridge/issues.lua#L375
 local function update_clock_metrics()
 
@@ -23,10 +25,11 @@ local function update_clock_metrics()
         end
     end
 
-    utils.set_gauge('clock_delta', 'Clock difference', min_delta * 1e-6, {delta = 'min'})
-    utils.set_gauge('clock_delta', 'Clock difference', max_delta * 1e-6, {delta = 'max'})
+    collectors_list.clock_delta = utils.set_gauge('clock_delta', 'Clock difference', min_delta * 1e-6, {delta = 'min'})
+    collectors_list.clock_delta = utils.set_gauge('clock_delta', 'Clock difference', max_delta * 1e-6, {delta = 'max'})
 end
 
 return {
-    update = update_clock_metrics
+    update = update_clock_metrics,
+    list = collectors_list,
 }
--- a/metrics/default_metrics/tarantool/init.lua
+++ b/metrics/default_metrics/tarantool/init.lua
@@ -1,23 +1,36 @@
 local metrics = require('metrics')
 
 local default_metrics = {
-    require('metrics.default_metrics.tarantool.network'),
-    require('metrics.default_metrics.tarantool.operations'),
-    require('metrics.default_metrics.tarantool.system'),
-    require('metrics.default_metrics.tarantool.replicas'),
-    require('metrics.default_metrics.tarantool.info'),
-    require('metrics.default_metrics.tarantool.slab'),
-    require('metrics.default_metrics.tarantool.runtime'),
-    require('metrics.default_metrics.tarantool.memory'),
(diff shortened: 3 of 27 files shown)


cartridge/roles/metrics.lua
```diff
@@ -138,6 +143,10 @@ local function apply_config(conf)
         end
         metrics_config_present = false
     end
+    metrics.clear()
+    metrics.enable_default_metrics(metrics_conf.include, metrics_conf.exclude)
+    metrics.enable_cartridge_metrics()
```
DifferentialOrange, line 148: Maybe cartridge metrics should be a section in default metrics so user will be able to include/exclude them too?

cartridge/roles/metrics.lua
```diff
@@ -138,6 +143,10 @@ local function apply_config(conf)
         end
         metrics_config_present = false
     end
+    metrics.clear()
```
DifferentialOrange, line 146: I can already see a tons of people (including myself) that got their custom metrics broken because of code design (like storing collector in a variable because searching it in O(N)-registry each time is painful). Moreover, what's about custom global labels and callbacks? It's not obvious how to write `clear()`-proof custom metrics code. I'm not sure that this is even possible. At least we never insisted on it and didn't have any examples and guides on how to do it. Casually adding this line of code will bring a great disaster

DifferentialOrange (changes requested):

metrics/default_metrics/tarantool/init.lua
```diff
@@ -1,23 +1,52 @@
...
+    end
+
+    for name, value in pairs(default_metrics) do
+        if #include > 0 then
+            if include_map[name] ~= nil then
+                metrics.register_callback(value.update)
+            else
+                metrics.unregister_callback(value.update)
+            end
+        elseif #exclude > 0 then
```
DifferentialOrange, line 41: This can be simplified to
```lua
if #include > 0 then
    if include_map[name] ~= nil then
        metrics.register_callback(value.update)
    else
        metrics.unregister_callback(value.update)
    end
elseif exclude[name] ~= nil then
    metrics.unregister_callback(value.update)
else
    metrics.register_callback(value.update)
end
```
but since your version is easier to read, you can ignore my comment

metrics/default_metrics/tarantool/init.lua
```diff
@@ -1,23 +1,52 @@
...
+    for _, name in ipairs(exclude) do
+        exclude_map[name] = true
+    end
+    local include_map = {}
+    for _, name in ipairs(include) do
+        include_map[name] = true
+    end
+
+    for name, value in pairs(default_metrics) do
+        if #include > 0 then
```
DifferentialOrange, line 35: Do you really should compute the length of each array every time instead of computing some flag like `linclude_mode/exclude_mode` above?

metrics/default_metrics/tarantool/init.lua
```diff
@@ -1,23 +1,52 @@
...
+    vinyl = require('metrics.tarantool.vinyl'),
+    luajit = require('metrics.tarantool.luajit'),
+    cartridge_issues = require('metrics.cartridge.issues'),
+    clock = require('metrics.cartridge.clock'),
 }
 
-local function enable()
-    for _, metric in ipairs(default_metrics) do
-        metrics.register_callback(metric.update)
+local function enable(include, exclude)
```
DifferentialOrange, line 21: Since one may use `metrics` without a role, shouldn't we validate "mixing include and exclude" case here too?

metrics/registry.lua
```diff
@@ -70,6 +70,14 @@ function Registry:register_callback(callback)
     end
 end
 
+function Registry:unregister_callback(callback)
+    for i, registered_callback in ipairs(self.callbacks) do
+        if registered_callback == callback then
```
DifferentialOrange, line 75: Isn't it a good time to make it a hash table like metrics registry?

DifferentialOrange: But since we don't have any explicit keys here, it is not so obvious how to organize it

test/integration/cartridge_role_test.lua
```diff
@@ -668,3 +670,81 @@ g.test_invalig_global_labels_names = function()
...
+    t.assert_equals(resp.status, 200)
+
+    local metrics_cnt = #resp.json
+    local vinyl_metrics = fun.iter(resp.json):filter(function(x)
+        return x.metric_name:find('tnt_vinyl')
+    end):length()
+    local lj_metrics = fun.iter(resp.json):filter(function(x)
+        return x.metric_name:find('lj_')
+    end):length()
+    local cartridge_metrics = fun.iter(resp.json):filter(function(x)
```
DifferentialOrange, line 700: I don't get it. `cartridge_issues` and `clock` metrics are disabled, isn't they?

test/integration/cartridge_role_test.lua
```diff
@@ -668,3 +670,81 @@ g.test_invalig_global_labels_names = function()
...
+                'vinyl', 'luajit',
+            }
+        }
+    })
+
+    local resp = server:http_request('get', '/metrics', {raise = false})
+    t.assert_equals(resp.status, 200)
+
+    local metrics_cnt = #resp.json
+    t.assert(metrics_cnt > 0)
```
DifferentialOrange, line 726: luatest 0.5.3 have compare asserts

test/integration/cartridge_role_test.lua
yngvar-antonsson: It's from previous version. Cartridge metrics count is 0 now. I'll delete it

test/integration/cartridge_role_test.lua
yngvar-antonsson: tnx

cartridge/roles/metrics.lua
```diff
@@ -46,11 +46,16 @@ local function set_labels(custom_labels)
...
-local function check_config(_)
+local function check_config(config)
     checks({
         export = 'table',
         ['global-labels'] = '?table',
+        include = '?table',
+        exclude = '?table',
     })
+    if config.include and config.exclude then
+        error("don't use exclude and include sections together", 0)
```
no1seman, line 57: Can't understand 2 things:
1. Why we need both include and exclude? One of the ways is to use only one single list
2. If use include and exclude  why not togather?

cartridge/roles/metrics.lua
```diff
@@ -153,6 +157,8 @@ local function apply_config(conf)
         end
     end
     apply_routes(paths)
+    metrics.enable_default_metrics(metrics_conf.include, metrics_conf.exclude)
```
no1seman, line 160: metrics.enable_default_metrics has no arguments?

metrics/registry.lua
```diff
@@ -70,6 +70,14 @@ function Registry:register_callback(callback)
...
+            table.remove(self.callbacks, i)
+        end
+    end
+end
+
 function Registry:set_labels(label_pairs)
     self.label_pairs = table.copy(label_pairs)
```
no1seman, line 82: may be label_pairs or {} since this method may be called straight from user code?

cartridge/roles/metrics.lua
```diff
@@ -144,7 +149,6 @@ end
 local function apply_config(conf)
     local metrics_conf = conf.metrics or {}
     metrics_conf.export = metrics_conf.export or {}
-    set_labels(metrics_conf['global-labels'])
```
rosik, line 147: Why is it necessary?

metrics/default_metrics/tarantool/init.lua
```diff
@@ -1,23 +1,63 @@
...
+    clock               = require('metrics.cartridge.clock'),
 }
 
-local function enable()
-    for _, metric in ipairs(default_metrics) do
-        metrics.register_callback(metric.update)
+local function delete_collectors(list)
+    for _, collector in pairs(list) do
+        metrics.registry:unregister(collector)
+    end
```
rosik, line 24:
```suggestion
    if list == nil then
        return
    end
    for _, collector in pairs(list) do
        metrics.registry:unregister(collector)
    end
    table.clear(list)
```

metrics/cartridge/clock.lua
```diff
@@ -1,5 +1,11 @@
 local utils = require('metrics.utils')
-local membership = require('membership')
+local ok, membership = pcall(require, 'membership')
+
+if not ok then
+    return { update = function() end }
```
rosik, line 5: setting default metrics will result in "attempt to index nil value 'list'" error

rosik: The fix is suggested below in `delete_collectors` function

metrics/default_metrics/tarantool/memory.lua
```diff
@@ -8,11 +10,13 @@ local function update_memory_metrics()
     if box.info.memory ~= nil then
         local i = box.info.memory()
         for k, v in pairs(i) do
-            utils.set_gauge('info_memory_' .. k, 'Memory' .. k, v)
+            local metric_name = 'info_memory_' .. k
+            collectors_list[metric_name] = utils.set_gauge(metric_name, 'Memory' .. k, v)
```
rosik, line 14:
```suggestion
            collectors_list[metric_name] = utils.set_gauge(metric_name, 'Memory ' .. k, v)
```

metrics/registry.lua
rosik: +1

metrics/tarantool/luajit.lua
```diff
@@ -13,6 +13,7 @@ end
 local function set_gauge(name, description, value, labels)
     local gauge = metrics.gauge(prefix_name(name), description)
     gauge:set(value, labels or {})
+    return gauge
 end
```
rosik, line 17: This is weird. Use `utils.set_gauge`instead of copy-pasting the code.

cartridge/roles/metrics.lua
yngvar-antonsson: Yes it should

cartridge/roles/metrics.lua
yngvar-antonsson: Removed `metrics.clear` call

metrics/default_metrics/tarantool/init.lua
yngvar-antonsson: Replaced it with next()

metrics/registry.lua
yngvar-antonsson: #276

cartridge/roles/metrics.lua
yngvar-antonsson: Yes

cartridge/roles/metrics.lua
yngvar-antonsson: 1) Yes, it's one possible way, but my point is to suggest users to choose: include only metrics they want, or include all metrics from default metrics list except metrics they don't need
2) That sections has separate logic. If you use both sections, which section should has higher priority?

metrics/registry.lua
yngvar-antonsson: That method not in metrics API and used only in `set_global_labels`.
https://github.com/tarantool/metrics/blob/9c41c9fcab21e048bf96da91d11e26ed183ee7e3/metrics/init.lua#L86

no1seman (approved):

metrics/tarantool/luajit.lua
yngvar-antonsson: #277

rosik (approved):

PR #287, opened 2021-08-16
Remove metrics by label

### LINKED ISSUE #263: Clean up metrics by label
It is possible to unregister a whole collector with registry internal methods. But it is not possible (at least without weird for-cycles and table removes) to unregister a specific metric value.

Let's consider the following case. We have a pool of connectors and several metrics of existing connectors (e.g. status, current pending requests, process successes and errors). If connector is deleted, it is no more needed to store its metrics. it would be a good pattern to use a single collector with different labels for a single metric, but then we wouldn't be able to clean it up if needed.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #263

### DIFF
--- a/metrics/collectors/histogram.lua
+++ b/metrics/collectors/histogram.lua
@@ -61,6 +61,18 @@ function Histogram:observe(num, label_pairs)
     end
 end
 
+function Histogram:remove(label_pairs)
+    assert(label_pairs, 'label pairs is a required parameter')
+    self.count_collector:remove(label_pairs)
+    self.sum_collector:remove(label_pairs)
+
+    for _, bucket in ipairs(self.buckets) do
+        local bkt_label_pairs = table.deepcopy(label_pairs)
+        bkt_label_pairs.le = bucket
+        self.bucket_collector:remove(bkt_label_pairs)
+    end
+end
+
 function Histogram:collect()
     local result = {}
     for _, obs in ipairs(self.count_collector:collect()) do
--- a/metrics/collectors/shared.lua
+++ b/metrics/collectors/shared.lua
@@ -14,6 +14,7 @@ function Shared:new_class(kind, method_names)
     table.insert(method_names, 'make_key')
     table.insert(method_names, 'append_global_labels')
     table.insert(method_names, 'collect')
+    table.insert(method_names, 'remove')
     local methods = {}
     for _, name in pairs(method_names) do
         methods[name] = self[name]
@@ -46,6 +47,13 @@ function Shared.make_key(label_pairs)
     end
     table.sort(parts)
     return table.concat(parts, '\t')
+end
+
+function Shared:remove(label_pairs)
+    assert(label_pairs, 'label pairs is a required parameter')
+    local key = self.make_key(label_pairs)
+    self.observations[key] = nil
+    self.label_pairs[key] = nil
 end
 
 function Shared:set(num, label_pairs)
--- a/metrics/collectors/summary.lua
+++ b/metrics/collectors/summary.lua
@@ -86,6 +86,16 @@ function Summary:observe(num, label_pairs)
     end
 end
 
+function Summary:remove(label_pairs)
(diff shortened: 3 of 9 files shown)


metrics/collectors/histogram.lua
```diff
@@ -61,6 +61,18 @@ function Histogram:observe(num, label_pairs)
     end
 end
 
+function Histogram:remove(label_pairs)
+    assert(label_pairs)
```
DifferentialOrange, line 65: How about something more verbose?

```
tarantool> assert(nil)
---
- error: assertion failed!
...
```
will definitely confuse users.

metrics/collectors/shared.lua
```diff
@@ -48,6 +49,13 @@ function Shared.make_key(label_pairs)
     return table.concat(parts, '\t')
 end
 
+function Shared:remove(label_pairs)
+    assert(label_pairs)
```
DifferentialOrange, line 53: How about something more verbose?

```
tarantool> assert(nil)
---
- error: assertion failed!
...
```
will definitely confuse users.

metrics/collectors/summary.lua
```diff
@@ -86,6 +86,16 @@ function Summary:observe(num, label_pairs)
     end
 end
 
+function Summary:remove(label_pairs)
+    assert(label_pairs)
```
DifferentialOrange, line 90: How about something more verbose?

```
tarantool> assert(nil)
---
- error: assertion failed!
...
```
will definitely confuse users.

test/collectors/shared_test.lua
```diff
@@ -16,3 +16,20 @@ g.test_different_order_in_label_pairs = function()
         {1, {a = 2, b = 1}},
     })
 end
+
+g.test_remove_metric_by_label = function()
+    local class = Shared:new_class('test_class', {'inc'})
+    local collector = class:new('test')
+    collector:inc(1, {a = 1, b = 2})
```
DifferentialOrange, line 23: Wait, is `{ string = number }` labels legal? Do export modules able to process it well?

It is not a question of this PR, maybe a ticket should be filed.

test/collectors_test.lua
```diff
@@ -268,3 +269,42 @@ g.test_global_labels = function()
...
+
+    local observations = c:collect()
+    t.assert_items_equals(fun.iter(observations):map(function(x) return {x.value, x.label_pairs} end):totable(), {
+        {1, {label = 1}},
+        {1, {label = 2}},
+    })
+
+    c:remove({label = 1})
+    observations = c:collect()
+    t.assert_items_equals(fun.iter(observations):map(function(x) return {x.value, x.label_pairs} end):totable(), {
```
DifferentialOrange, line 288: Why not use `assert_observations`?

DifferentialOrange (commented):
I left a couple of comments. PR seems legit

test/collectors/shared_test.lua
DifferentialOrange: Refer to https://github.com/tarantool/metrics/issues/289

metrics/collectors/histogram.lua
```diff
@@ -61,6 +61,18 @@ function Histogram:observe(num, label_pairs)
     end
 end
 
+function Histogram:remove(label_pairs)
+    assert(label_pairs, 'label pairs is required parameter')
```
DifferentialOrange, line 65:
```suggestion
    assert(label_pairs, 'label pairs is a required parameter')
```

metrics/collectors/histogram.lua
```diff
@@ -61,6 +61,18 @@ function Histogram:observe(num, label_pairs)
     end
 end
 
+function Histogram:remove(label_pairs)
+    assert(label_pairs, 'label pairs is a required parameter')
```
DifferentialOrange, line 65:
```suggestion
    assert(label_pairs, 'label_pairs is a required parameter')
```

DifferentialOrange (approved):

PR #288, opened 2021-08-16
Add counter:reset method

### LINKED ISSUE #260: Counter initialization
It is implied that counter start from 0, since counter value after its first `inc(1)` will be `1`.

But metric value do not exist until first `inc()` call (since there are no other methods to change counter metric value). So if value if never changed, metric will be null.

For example, it is a common case to count process errors with counter. But if we did not have any errors from the start, the metric will be null. HTTP activity is the same (https://github.com/tarantool/metrics/issues/96): it is more convenient to know that we had `0` requests on some route rather then `no data`. `No data` if often treated like "there is something wrong with metric retrieval" rather then "it's a plain zero".

It is now possible to initialize counter with specific labels with
```lua
counter:inc(0, label_pairs)
```
but it is definitely a duct tape. It would be better to support this behavior explicitly, e.g. with some collector method.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #260

### DIFF
--- a/metrics/collectors/counter.lua
+++ b/metrics/collectors/counter.lua
@@ -9,4 +9,8 @@ function Counter:inc(num, label_pairs)
     Shared.inc(self, num, label_pairs)
 end
 
+function Counter:reset(label_pairs)
+    Shared.set(self, 0, label_pairs)
+end
+
 return Counter
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 ### Added
 - collector's method `remove` to clear observations with given label pairs [#263](https://github.com/tarantool/metrics/issues/263)
+- `counter:reset()` method [#260](https://github.com/tarantool/metrics/issues/260)
 
 ## [0.10.0] - 2021-08-03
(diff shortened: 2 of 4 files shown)


metrics/collectors/counter.lua
```diff
@@ -9,4 +9,8 @@ function Counter:inc(num, label_pairs)
     Shared.inc(self, num, label_pairs)
 end
 
+function Counter:reset(label_pairs)
```
DifferentialOrange, line 12: The original request was about counter initialization. Initialization with `counter:reset(label_pairs)` seems kinda weird. How about adding an alias for `reset` as a little syntactic sugar?

```lua
Counter.initialize = reset
```

DifferentialOrange (commented):
PR's missing changelog and doc edits

metrics/collectors/counter.lua
rosik: init / reinit / reset, who cares. Reset is better than inc(0), and that's the purpose.

rosik: Init which resets the value may be considered weird too.

DifferentialOrange (approved):
Well, I'm still prefer having both `init` and `reset`, but if you're against it, let's at least add the functionality

PR #291, opened 2021-08-23
Refactor metrics.utils and change type of luajit metrics

### LINKED ISSUE #277: Refactor metrics.utils to avoid duplicate code
This is weird. Use `utils.set_gauge`instead of copy-pasting the code.

_Originally posted by @rosik in https://github.com/tarantool/metrics/pull/268#discussion_r680089726_

yngvar-antonsson:
Luajit metrics should be gauge, because they're set in Tarantool, not metrics module.
I also refactor metrics.utils module and remove unused internal function `prefix_name ` from it's API.

@DifferentialOrange, will it affect Grafana dashboard?

I didn't forget about

- [ ] Changelog

Close #277


metrics/tarantool/luajit.lua
```diff
@@ -1,30 +1,9 @@
-local metrics = require('metrics')
```
olegrok, line 1: I hope you considered this thread when push this changes https://github.com/tarantool/metrics/pull/128#discussion_r505303306

metrics/tarantool/luajit.lua
```diff
@@ -34,46 +13,52 @@ local function update()
     -- Details: https://github.com/tarantool/doc/issues/1597
     local lj_metrics = misc.getmetrics()
     collectors_list.gc_freed =
-        set_counter('gc_freed', 'Total amount of freed memory', lj_metrics.gc_freed)
+        utils.set_gauge('gc_freed', 'Total amount of freed memory', lj_metrics.gc_freed, nil, LJ_PREFIX)
```
DifferentialOrange, line 16: There was some ideas on `reset`/`inc` counter approach for monotonic metrics here: https://github.com/tarantool/metrics/pull/128#discussion_r508520625 . Maybe we should use this approach when https://github.com/tarantool/metrics/pull/288#pullrequestreview-737155916 instead of the current one? For example, when using prometheus type of metric is shown in Grafana UI.

PR #292, opened 2021-08-23
Add note about that all numbers in label_pairs are treated as strings

### LINKED ISSUE #289: Non-string labels may be confusable
We permit to use every concatenatable type as a `label_pairs` key or value (except for global labels, which (asserts that key is a string)[https://github.com/tarantool/metrics/blob/91629f74db79cd6fb839b5d340dc49f9e186a570/metrics/init.lua#L95-L97]). It may introduce some unexpected behavior like the following.

```lua
tarantool> c = metrics.counter('c')                                         
---
...

tarantool> c:inc(1, {l = '1'})     
---
...

tarantool> metrics.collect()       
---
- - label_pairs:
      l: '1'
    timestamp: 1629275058158257
    value: 1
    metric_name: c
...

tarantool> c:inc(1, {l = 1})       
---
...

tarantool> metrics.collect()
---
- - label_pairs:
      l: 1
    timestamp: 1629275064319510
    value: 2
    metric_name: c
...
```
Notice that `label_pairs` content is changing (it will not change the output of collectors since they convert labels to string).

As far as I know, Prometheus store all labels as strings. json format that is used by Telegraf has similar behavior. I think it should be discussed if number labels is legal (for example, histogram buckets in our code [defined as numbers](https://github.com/tarantool/metrics/blob/91629f74db79cd6fb839b5d340dc49f9e186a570/metrics/collectors/histogram.lua#L5-L6) and [used as labels](https://github.com/tarantool/metrics/blob/91629f74db79cd6fb839b5d340dc49f9e186a570/metrics/collectors/histogram.lua#L52)). If we decide to leave this behavior as it is, it should be at least mentioned in the documentation that all numbers are treated as strings.

yngvar-antonsson:
I didn't forget about

Close #289


doc/monitoring/api_reference.rst
```diff
@@ -46,7 +46,8 @@ Counter
 
         :param number        num: Increase value.
         :param table label_pairs: Table containing label names as keys,
-                                  label values as values.
+                                  label values as values. Note that all
+                                  numbers in label_pairs are treated as strings.
```
DifferentialOrange, line 50:
```suggestion
                                  keys in label_pairs are treated as strings.
```

DifferentialOrange: I think it's closer to the truth

yngvar-antonsson: Maybe 'label values'?
Keys sound like "label names"

DifferentialOrange: Isn't it both keys and values in table pairs? They both concatenated on metric observation as strings and casted to strings in both json and Prometheus export

yngvar-antonsson: Maybe "Note that **both labels names and values** in label_pairs are treated as strings" ?

DifferentialOrange: Seems legit

DifferentialOrange:
Are you sure that we should not treat the following situation where label_pairs content is changing as a bug?
```
tarantool> c:inc(1, {l = '1'})     
---
...

tarantool> metrics.collect()       
---
- - label_pairs:
      l: '1'
    timestamp: 1629275058158257
    value: 1
    metric_name: c
...

tarantool> c:inc(1, {l = 1})       
---
...

tarantool> metrics.collect()
---
- - label_pairs:
      l: 1
    timestamp: 1629275064319510
    value: 2
    metric_name: c
...
```

I still fear that users with custom collectors (if there's any) or possible future collectors in this package may distinct strings from numbers.

yngvar-antonsson:
@DifferentialOrange 
1) No one reported this yet 
2) And if we change behavior in new release, it would break something in users code
3) So I propose to mention it in docs and don't change

DifferentialOrange:
> @DifferentialOrange
> 
>     1. No one reported this yet
> 
>     2. And if we change behavior in new release, it would break something in users code
> 
>     3. So I propose to mention it in docs and don't change

Well, if you say so

DifferentialOrange (approved):

PR #294, opened 2021-08-25
Add event loop metric

### LINKED ISSUE #123: Tarantool tx thread event loop speed
# Motivation

In case when there is some heavy fiber or fibers amount is big it's possible that fiber takes control too late.

For e.g. if fiber has some timeout logic, timeout can be triggered even if necessary data is arrived in buffer.

for e.g.

```lua
while clock.now() < deadline do
  local payload = sock.read(100, timeout=0.1)
end
-- payload is ignored
if clock.now() >= deadline then
  error
end
```

## Propolsal

Measure tx thread event loop speed with special fiber

```lua
require('strict')
local log = require('log')
local clock = require('clock')
local fiber = require('fiber')

local timeout = 1

local tx_loop_delay_gauge = metrics.gauge('tnt_tx_loop_delay', 'Tarantool tx thread event loop delay')
if rawget(_G, '__metrics_tx_speed') == nil then
    _G.__metrics_tx_speed = fiber.new(function ()
            fiber.self():name("tx_speed")

            while true do
                local start = clock.monotonic64()
                fiber.sleep(tonumber(timeout))
                local diff = clock.monotonic64() - start
                local err = (tonumber(diff) - timeout*1e9)/1e9

                if err < 0 then
                    --log.info("Early wakeup %f", err)
                elseif err > 0 then
                    --log.info('Event loop slower by %f seconds', err)
                end
                tx_loop_delay_gauge:set(timeout + err)
            end
    end)
end
```

yngvar-antonsson:
What has been done? Why? What problem is being solved?

I didn't forget about

- [x] Tests
- [ ] Changelog
- [x] Rockspec and rpm spec

Close #123

### DIFF
--- a/metrics/tarantool/event_loop.lua
+++ b/metrics/tarantool/event_loop.lua
@@ -1,0 +1,43 @@
+local clock = require('clock')
+local fiber = require('fiber')
+local metrics = require('metrics')
+
+local collectors_list = {}
+
+local function monitor()
+    fiber.self():name("tx_speed")
+
+    local max = -math.huge
+
+    while true do
+        local start = clock.monotonic64()
+        fiber.yield()
+        local diff = clock.monotonic64() - start
+        -- measure max time between in case
+        -- if maximum diff was between two metrics.collect() calls
+        if diff > max then
+            max = diff
+        end
+        local collector = metrics.registry:find('gauge', 'tnt_tx_loop_delay')
+        if not collector then
+            -- if collector was removed from registry
+            rawset(_G, '__metrics_tx_speed', nil)
+            return
+        end
+        collector:set(tonumber(diff)/10^6, {time = 'current'})
+        collector:set(tonumber(max)/10^6, {time = 'max'})
+    end
+end
+
+
+local function update()
+    collectors_list.tnt_tx_loop_delay = metrics.gauge('tnt_tx_loop_delay', 'Tarantool tx thread event loop delay')
+    if rawget(_G, '__metrics_tx_speed') == nil then
+        rawset(_G, '__metrics_tx_speed', fiber.create(monitor))
+    end
+end
+
+return {
+    update = update,
+    list = collectors_list,
+}
--- a/metrics-scm-1.rockspec
+++ b/metrics-scm-1.rockspec
@@ -54,6 +54,7 @@ build = {
         ['metrics.psutils.psutils_linux']                = 'metrics/psutils/psutils_linux.lua',
         ['metrics.tarantool.luajit']                     = 'metrics/tarantool/luajit.lua',
         ['metrics.tarantool.vinyl']                      = 'metrics/tarantool/vinyl.lua',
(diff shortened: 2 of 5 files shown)


metrics/tarantool/event_loop.lua
```diff
@@ -0,0 +1,43 @@
...
+local metrics = require('metrics')
+
+local collectors_list = {}
+
+local function monitor()
+    fiber.self():name("tx_speed")
+
+    local max = -math.huge
+
+    while true do
```
filonenko-mikhail, line 12: Please avoid `while true` pattern as aggressive snippet.
Detox it with something else

@ochaton example

https://gist.github.com/ochaton/6bc9a7030df7b4e7c63703599a5b2f96

metrics/tarantool/event_loop.lua
```diff
@@ -0,0 +1,43 @@
...
+        local start = clock.monotonic64()
+        fiber.yield()
```
ochaton, line 14: Firstly, monotonic64 doesn't show nearly realtime. It may produce times which a not related to real wallclock.

It seems that ev_time (fiber.time()) is actually uses simple clock_gettime
https://github.com/enki/libev/blob/master/ev.c#L1854

Secondly, it is too bad to run this loop without fiber.sleep(interval), because now this fiber will be constantly scheduled in each ev_once.

It is good to notice, that diff will be measured at the beginning of the single run, while start is measured nearly at any point in RUN (but in this code also in the beginning of the loop).

I suggest you to write something like this:
```lua
fiber.sleep(interval) -- sleep interval, and this will guarantee us that next line will be executed nearly at the beginning of the loop (after some timers)
local wall_1, ev_1 = clock.realtime64(), fiber.time64()*1e3
fiber.sleep(0) -- reset watcher to EV, to reach next loop. Actually EV has ev_postpone, but we do not have such API in Lua.
local wall_2, ev_2 = clock.realtime64(), fiber.time64()*1e3

-- Do your mesurments: you will have 3 intervals:
-- (if everything is ok with ntp):
-- ev_1 <= wall_1 <= ev_2 <= wall_2
-- ev_loop_time is ev_2 - ev_1 (it is mesured by the EV itself at the beginning of the loop)
-- wall_2 - wall_1 -- is actual sleep of the fiber. It does mean not that much actually
-- wall_2 - ev_2 -- is lag from start of this run of loop (if it is large, then we might have some long code in timers)
-- ev_2 - wall_1 -- is lag of "epilog" of the loop. If it is large then exists fibers which are running long while having fiber.sleep(timeout > 0) or long requests
```

Read more info about my experiments: https://gist.github.com/ochaton/6bc9a7030df7b4e7c63703599a5b2f96

Also you have to provide mechanism to stop your monitoring fiber and maybe get collector before while-loop

ochaton (changes requested):

filonenko-mikhail:
gently closing

PR #296, opened 2021-09-08
Throw an error when http_middelware is processing a wrong handler

### LINKED ISSUE #199: Process latency count more carefully
If user forgets to return correct response object in HTTP handler, it results in "trying to subscribe `nil`" error in this line:
https://github.com/tarantool/metrics/blob/a2d783e5b68475a7c88f89b5ea10ce470d8e67fa/metrics/http_middleware.lua#L75

We should process this more carefully. For example, throw explicit `error("incorrect http handler")`.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #199

### DIFF
--- a/metrics/http_middleware.lua
+++ b/metrics/http_middleware.lua
@@ -71,8 +71,9 @@ end
 -- ... arguments for pcall to instrument
 function export.observe(collector, route, ...)
     return collector:observe_latency(function(ok, result)
-        if result == nil then
-            error("incorrect http handler", 0)
+        if type(result) ~= 'table' then
+            error(('incorrect http handler for %s %s: expecting return response object'):
+                format(route.method, route.path), 0)
         end
         return {
             path = route.path,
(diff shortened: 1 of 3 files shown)


metrics/http_middleware.lua
```diff
@@ -71,6 +71,9 @@ end
 -- ... arguments for pcall to instrument
 function export.observe(collector, route, ...)
     return collector:observe_latency(function(ok, result)
+        if result == nil then
```
DifferentialOrange, line 74: non-table result will fail with `trying to subscribe a string value`, so I think `type(result) == 'table'` is the condition.

metrics/http_middleware.lua
```diff
@@ -71,6 +71,9 @@ end
...
+            error("incorrect http handler", 0)
```
DifferentialOrange, line 75: I think it should be more explicit, like `expecting {status = status, ...}` return

DifferentialOrange (changes requested):

metrics/http_middleware.lua
yngvar-antonsson: But status can be omitted: 

```lua
status = (not ok and 500) or result.status or 200,
```

DifferentialOrange (approved):

PR #299, opened 2021-09-09
Fix cartridge issues metric call before cartridge.cfg()

### LINKED ISSUE #298: Stateboard instance fails to collect cartridge issues
After migration from metrics 0.9.0 to 0.10.0 and from cartridge 2.6.0 to 2.7.1 tests in our project started to fail.

We've got stateboard with default metrics and tests on this metrics.

```lua
-- stateboard,init.lua
metrics.enable_default_metrics()
metrics.set_global_labels({
    alias = app_config.alias
})
--set http endpoint
```

Test goes like this
```lua
group.test_stateboard_metrics = function()
    local resp = stateboard:http_request('get', '/app/metrics')
    test.assert_not(next(resp.json) == nil, "Metrics response must not be empty")
    for _, metric in ipairs(resp.json) do
        test.assert_equals(metric.label_pairs.alias, stateboard.env.TARANTOOL_ALIAS)
    end
end
```

After package update we've started to get an internal error
```
1) {
     response = {
        body = "Unhandled error: ...app/.rocks/share/tarantool/cartridge/issues.lua:97: attempt to index local 'topology_cfg' (a nil value)
                     stack traceback:
                     ...app/.rocks/share/tarantool/http/server.lua:743: in function 'process_client'
                     ...app/.rocks/share/tarantool/http/server.lua:1199: in function <..app/.rocks/share/ефкфтещщдhttp/server.lua:1198>
                     [C]: in function 'pcall'
                     builtin/socket.lua:1058: in function <builtin/socket.lua:1056>
                     
                     Request:
                     get /app/metrics? HTTP/1.1
                     Host: localhost:8088
                     Content-type: application/x-www-form-urlencoded
                     Connection: close
                     Content-length: 0
                     Accept: */*
                     
                     ",
             headers = {
                  connection = "close",
                  ["content-length"] = "667",
                  server = "Tarantool http (*** v1.10.11-0-gf0b0e7ecf-r423)",
             },
             proto = {1, 1},
             reason = "Ok",
             status = 500,
     },
       type = "HTTPRequest",
}
```

After I changed stateboard,init.lua to exclude `cartridge_issues`, it has been fixed.
```
metrics.enable_default_metrics(
    nil,                   --include section
    { 'cartridge_issues' } --exclude section
)
```

Even if this fix is reasonable and satisfying, I think this error throw is a bug and it must be fixed. It may be not `metrics`, but `cartridge` bug.

P.S. I can DM you a link to our repo project.

yngvar-antonsson:
Before `cartridge.cfg()` call `list_on_instance` throw an error. I pcall it.

I didn't forget about

- [x] Tests
- [x] Changelog

Close #298

### DIFF
--- a/metrics/cartridge/issues.lua
+++ b/metrics/cartridge/issues.lua
@@ -3,14 +3,18 @@ local fun = require('fun')
 
 local collectors_list = {}
 
-local function update_info_metrics()
+local function update()
     local list_on_instance = rawget(_G, '__cartridge_issues_list_on_instance')
 
     if not list_on_instance then
         return
     end
 
-    local issues = list_on_instance()
+    local ok, issues = pcall(list_on_instance)
+
+    if not ok then
+        return
+    end
 
     local levels = { 'warning', 'critical' }
 
@@ -22,6 +26,6 @@ local function update_info_metrics()
 end
 
 return {
-    update = update_info_metrics,
+    update = update,
     list = collectors_list,
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 - Throw an error when http_middelware is processing a wrong handler [#199](https://github.com/tarantool/metrics/issues/199)
+- cartridge issues metric fails before cartridge.cfg() call [#298](https://github.com/tarantool/metrics/issues/298)
 
 ## [0.10.0] - 2021-08-03
 ### Changed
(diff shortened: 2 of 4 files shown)


metrics/cartridge/issues.lua
```diff
@@ -3,14 +3,18 @@ local fun = require('fun')
...
 
     if not list_on_instance then
         return
     end
 
-    local issues = list_on_instance()
+    local ok, issues = pcall(list_on_instance)
+
+    if not ok then
+        return
```
DifferentialOrange, line 16: What about some log.error message?

DifferentialOrange (approved):

metrics/cartridge/issues.lua
yngvar-antonsson: In case of stateboard it will log error each time metric will be collected, I don't think this is appropriate

PR #300, opened 2021-09-09
Add read_only metric

### LINKED ISSUE #275: Add master/slave history (faliover) metric
It wiil be great to add one simple metric that will return box.info.ro==false => 1 box.info.ro== true> 0 on each metrics scrape. This will give us an option to draw a graph of history of master switching over all instances of cluster and then we can easily filter instances by replicasets (may be for this only metric add replicaset alias label for Cartridge apps and replicaset UUID for non-Cartridge ones?).

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog
- [x] Picture from Grafana

Close #275


DifferentialOrange:
Possible Grafana visualizations
image
image

no1seman (approved):

PR #310, opened 2021-09-24
Expand documentation 

### LINKED ISSUE #111: Expand documentation on http_middleware module
It is not clear from an [example](https://github.com/tarantool/metrics/blob/master/example/HTTP/latency_v1.lua), that in order for this to work correctly you need ONLY ONE collector for router. Preferably even default collector.

Otherwise you get inconsistent metric names and a lot of metrics, like: 
```
  {
    "label_pairs": {
      "path": "/labels",
      "method": "GET",
      "status": 200,
      "alias": "tnt-router"
    },
    "timestamp": 1600351181350306,
    "metric_name": "labels_latency_avg",
    "value": 0
  },
```
instead of:
```
  {
    "label_pairs": {
      "path": "/labels",
      "method": "GET",
      "status": 200,
      "alias": "tnt-router"
    },
    "timestamp": 1600352083432399,
    "metric_name": "http_server_request_latency_avg",
    "value": 0
  },
```

So when registering routes for endpoints, everything must use the same collector. This can be done using default collector. Something along the lines of:
```
local http_middleware = require('cartridge').service_get('metrics').http_middleware
if http_collector == nil then
    http_collector = http_middleware.build_default_collector('average')
end

server:route({ path = "/labels", method = "ANY"}, http_middleware.v1(handler, http_collector))
```

### LINKED ISSUE #264: Best practices
Metrics documentation lacks best practices section. API doc is good, but it is not obvious how you should use collectors and callbacks in real life.

For example, last time someone asked me about how to set up HTTP middleware I sent him to read [my article](https://habr.com/ru/company/mailru/blog/534826/) since I wasn't be able to find a relevant example in documentation. How should one use callbacks in a role? What's the best way to create/delete a collector for specific object, if we use OOP in Lua? We should find the answers to this questions and write them down.

yngvar-antonsson:

Close #111 
Close #264

### DIFF
--- a/doc/monitoring/api_reference.rst
+++ b/doc/monitoring/api_reference.rst
@@ -343,9 +343,41 @@ Metrics functions
     some global label, the method argument value will be used.
 
     Note that both labels names and values in label_pairs are treated as strings.
+
 ..  function:: collect()
 
     Collect observations from each collector.
+
+..  class:: registry
+
+    ..  method:: unregister(collector)
+
+        Removes collector from registry
+
+        :param table collector: collector that has to be removed
+
+        Example:
+
+        ..  code-block:: lua
+            local collector = metrics.gauge('some-gauge')
+
+            -- after some time we don't need it anymore
+
+            metrics.registry:unregister(collector)
+
+    ..  method:: find(kind, name)
+
+        Finds collector in registry
+
+        :param string kind: collector kind ('counter', 'gauge', 'histogram' or 'summary')
+        :param string name: collector name
+
+        Example:
+
+        ..  code-block:: lua
+            local collector = metrics.gauge('some-gauge')
+
+            collector = metrics.registry:find('gauge', 'some-gauge')
 
 ..  function:: register_callback(callback)
 
@@ -356,6 +388,15 @@ Metrics functions
 
     Most common usage is for gauge metrics updates.
 
+    Example:
+
+    ..  code-block:: lua
+
+        metrics.register_callback(function()
+            local cpu_metrics = require('metrics.psutils.cpu')
+            cpu_metrics.update()
+        end)
+
 ..  function:: unregister_callback(callback)
 
     Unregisters a function ``callback`` which will be called right before metrics
@@ -364,6 +405,26 @@ Metrics functions
     :param function callback: Function which takes no parameters.
 
     Most common usage is for unregister enabled callbacks.
+
+    Example:
+
+    ..  code-block:: lua
+
+        local cpu_callback = function()
+            local cpu_metrics = require('metrics.psutils.cpu')
+            cpu_metrics.update()
+        end
+
+        metrics.register_callback(cpu_callback)
+
+        -- after some time we dont need that callback anymore
+
+        metrics.unregister_callback(cpu_callback)
+
+..  function: invoke_callbacks()
+
+    Invokes all registered callbacks. Needs to be called before each export.
+    If youre using one of default exporters, invoke_callbacks function will be called by exporter
 
 .. _metrics-role-functions:
 
--- a/doc/monitoring/getting_started.rst
+++ b/doc/monitoring/getting_started.rst
@@ -35,12 +35,6 @@ Enable default Tarantool metrics such as network, memory, operations, etc:
 
     metrics.enable_default_metrics()
 
-If you use Cartridge, enable Cartridge metrics:
-
-.. code-block:: lua
-
-    metrics.enable_cartridge_metrics()
-
 Initialize the Prometheus Exporter, or export metrics in any other format:
 
 .. code-block:: lua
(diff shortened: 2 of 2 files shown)


doc/monitoring/api_reference.rst
```diff
@@ -365,6 +406,26 @@ Metrics functions
 
     Most common usage is for unregister enabled callbacks.
```
DifferentialOrange, line 407: Well, this is the only one usage

doc/monitoring/api_reference.rst
```diff
@@ -365,6 +406,26 @@ Metrics functions
...
+
+        metrics.register_callback(cpu_callback)
+
+        -- after some time we dont need that callback anymore
+
+        metrics.unregister_callback(cpu_callback)
+
+..  function: invoke_callbacks()
+
+    Invokes all registered callbacks. Needs to be called before each export.
```
DifferentialOrange, line 426: Maybe
```suggestion
    Invokes all registered callbacks. Needs to be called before each collect.
```

doc/monitoring/getting_started.rst
```diff
@@ -63,6 +57,55 @@ Now you can use the HTTP API endpoint ``/metrics`` to collect your metrics
...
+
+    -- Set router summary latency collection middleware
+    router:use(metrics.http_middleware.v2(collector), { name = 'latency_instrumentation' })
+
+    -- Start HTTP routing using configured router
+    httpd:set_router(router)
+    httpd:start()
+
+Note that you need only one collector to collect all http metrics.
+If youre using default Grafana-dashboard (link) dont change collector name and description,
```
DifferentialOrange, line 105: Only collector name, description change will not affect anything

doc/monitoring/getting_started.rst
```diff
@@ -63,6 +57,55 @@ Now you can use the HTTP API endpoint ``/metrics`` to collect your metrics
...
+
+    -- Set route handler with summary latency collection
+    httpd:route({ path = '/path-1', method = 'POST' }, metrics.http_middleware.v1(handler_1, collector))
+    httpd:route({ path = '/path-2', method = 'GET' }, metrics.http_middleware.v1(handler_2, collector))
+
+    -- Start HTTP routing
+    httpd:start()
+
+Using HTTP v2:
+
```
DifferentialOrange, line 84: extra line

doc/monitoring/getting_started.rst
```diff
@@ -63,6 +57,55 @@ Now you can use the HTTP API endpoint ``/metrics`` to collect your metrics
...
+.. code-block:: lua
+
+    local httpd = require('http.server').new(ip, port)
+
+    -- Create summary latency collector
+    local collector = metrics.http_middleware.build_default_collector('summary')
+
+    -- Set route handler with summary latency collection
+    httpd:route({ path = '/path-1', method = 'POST' }, metrics.http_middleware.v1(handler_1, collector))
```
DifferentialOrange, line 77: Well, since we recommend to use the only one collector, maybe it is better to give an example with `configure_default_collector` as a basic approach and do not pass any arguments explicitly?

doc/monitoring/api_reference.rst
```diff
@@ -343,10 +343,42 @@ Metrics functions
...
 
     Collect observations from each collector.
 
+..  class:: registry
+
+    ..  method:: unregister(collector)
+
+        Removes collector from registry
+
+        :param table collector: collector that has to be removed
```
DifferentialOrange, line 357: Maybe it should be specified that it is a collector object/collector table

doc/monitoring/api_reference.rst
```diff
@@ -343,10 +343,42 @@ Metrics functions
...
+        Example:
+
+        ..  code-block:: lua
+            local collector = metrics.gauge('some-gauge')
+
+            -- after some time we don't need it anymore
+
+            metrics.registry:unregister(collector)
+
+    ..  method:: find(kind, name)
```
DifferentialOrange, line 368: Possible return values? It is obvious that in case of success it returns a collector, but what in case of an error?

doc/monitoring/getting_started.rst
yngvar-antonsson: default_collector in `configure_default_collector` is `histogram`

DifferentialOrange: `build_default_collector` is a confusing name itself, since it is a plain `build_collector` and do not have any deal with setting up defaults. This code is confusing as well. I think we shouldn't use `build_default_collector` in examples and should recommend the following approach: if you do not want to use default histogram collector, use `configure_default_collector('summary', params)` in your code.

DifferentialOrange (changes requested):

doc/monitoring/getting_started.rst
patiencedaur: Hi @yngvar-antonsson, @DifferentialOrange. I am in charge of proofreading and localizing Tarantool docs. Kindly let me know whether you're going to introduce more changes to the metrics documentation. Thank you.

NickVolynkin: Hi folks. Good enough docs right now are better than perfect docs someday. Let's just merge it as is and make one more iteration someday later.

PR #317, opened 2021-10-29
Deprecate HTTP v2 middleware

yngvar-antonsson:
Since HTTP v2 was deprecated, we deprecate middleware v2.


metrics/http_middleware.lua
```diff
@@ -103,6 +104,7 @@ end
 -- @return middleware
 -- @usage router:use(http_middleware.v2(), {name = 'http_instrumentation'})
 function export.v2(collector)
+    log.warn('HTTP v2 is deprecated and will be removed in next releases. Use HTTP v1 instead')
```
olegrok, line 107: You can't write in log on each request. It should be done only at first request to avoid unexpected (for user) log flooding.

olegrok: See https://github.com/tarantool/errors/blob/master/errors/deprecate.lua or https://github.com/tarantool/vshard/commit/b035fd44fd02b0feb22cb313a208328026ec8d3e#diff-85f9e399cd9ae941a0d5d57372aa49892d5e2798364c02ddc335f6af08060709R1066

olegrok (approved):

Totktonada:
It seems, it worth to drop http v2 example from the Prometheus plugin documentation: https://github.com/tarantool/metrics/blob/master/doc/monitoring/plugins.rst (line 54-63 and 90-101).

Totktonada:
Filed as #324 to don't miss it.

I was pointed to this part of the documentation by a user, which was confused by all those v1 / v2 perturbations.

PR #322, opened 2021-11-03
Cast number64 to json number in export plugin

### LINKED ISSUE #321: Json values with ULL suffix
Some metrics parse their values to ULL strings:
```json
[{
   "label_pairs":{
      "engine":"memtx",
      "name":"MY_SPACE",
      "alias":"tnt_router"
   },
   "timestamp":1635867200499883,
   "metric_name":"tnt_space_bsize",
   "value":"51589ULL"
},
{
   "label_pairs":{
      "engine":"memtx",
      "name":"MY_SPACE",
      "alias":"tnt_router"
   },
   "timestamp":1635867200499883,
   "metric_name":"tnt_space_total_bsize",
   "value":"199045ULL"
}]
```
This values can't be parsed by Telegraf+InfluxDB monitoring stack. This issue was already solved for Prometheus (https://github.com/tarantool/metrics/issues/15) and Graphite (https://github.com/tarantool/metrics/issues/90), but not for json.

DifferentialOrange:
Closes #321

### DIFF
--- a/metrics/plugins/json.lua
+++ b/metrics/plugins/json.lua
@@ -6,6 +6,8 @@ local function finite(value)
     if type(value) == "string" then
         value = tonumber(value)
         if value == nil then return nil end
+    elseif type(value) == "cdata" then -- support number64
+        return value
     elseif type(value) ~= "number" then
         return nil
     end
--- a/test/plugins/json_test.lua
+++ b/test/plugins/json_test.lua
@@ -58,3 +58,23 @@ g.test_number_value_serialization = function()
     t.assert_equals(float_obs.value, 0.333, 'check float')
     t.assert_equals(int_obs.value, 10, 'check int')
 end
+
+g.test_number64_ll_value_parses_to_json_number = function()
+    local gauge_ll = metrics.gauge('test_cdata_ll')
+    gauge_ll:set(-9007199254740992LL)
+
+    local json_metrics = json.decode(json_exporter.export())
+    local obs_ll = utils.find_obs('test_cdata_ll', {}, json_metrics)
+    t.assert_not_equals(type(obs_ll.value), 'string', 'number64 is not casted to string on export')
+    t.assert_equals(obs_ll.value, -9007199254740992LL, 'number64 LL parsed to corrent number value')
+end
+
+g.test_number64_ull_value_parses_to_json_number = function()
+    local gauge_ull = metrics.gauge('test_cdata_ull')
+    gauge_ull:set(9007199254740992ULL)
+
+    local json_metrics = json.decode(json_exporter.export())
+    local obs_ull = utils.find_obs('test_cdata_ull', {}, json_metrics)
+    t.assert_not_equals(type(obs_ull.value), 'string', 'number64 is not casted to string on export')
+    t.assert_equals(obs_ull.value, 9007199254740992ULL, 'number64 ULL parsed to corrent number value')
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Update `http` dependency to 1.1.1
 
+### Fixed
+- Cast number64 to json number in json export plugin [#321](https://github.com/tarantool/metrics/issues/321)
+
 ### Deprecated
 - HTTP middleware v2
 


metrics/plugins/json.lua
```diff
@@ -6,6 +6,11 @@ local function finite(value)
     if type(value) == "string" then
         value = tonumber(value)
         if value == nil then return nil end
+    elseif type(value) == "cdata" then
+        value = tostring(value)
+        -- Luajit cdata type inserts some postfix in the end of the number after tostring() operation
+        value = value:gsub("U*LL", "")
```
Totktonada, line 12: Nit: we can use `U?LL` here.

metrics/plugins/json.lua
```diff
@@ -6,6 +6,11 @@ local function finite(value)
...
+        return tonumber(value)
```
Totktonada, line 13: How about numbers out of the range [-2^53; 2^53] range, where double (lua number) types cannot represent all integral numbers? There is `number64()` function. BTW, it returns just `number` if the value is small enough.

(I'm not in context here, but, it seems, this part of the code may be just `number64()` call.)

test/plugins/json_test.lua
```diff
@@ -58,3 +58,21 @@ g.test_number_value_serialization = function()
     t.assert_equals(float_obs.value, 0.333, 'check float')
     t.assert_equals(int_obs.value, 10, 'check int')
 end
+
+g.test_cdata_ll_value_parses_to_number = function()
+    local gauge_ll = metrics.gauge('test_cdata_ll')
+    gauge_ll:set(-205LL)
```
Totktonada, line 64: number64 values (cdata<int64_t> / cdata<uint64_t>) usually occurs when you working with very large (or very small negative) integral numbers. So I think that we should test those cases.

DifferentialOrange: Well, I've got ULL suffix in `51589ULL`. But I can increase the number

metrics/plugins/json.lua
DifferentialOrange: Oh, thank you. I've based my solution on https://github.com/tarantool/metrics/pull/23 , but I think it is worth to use more simpler one

test/plugins/json_test.lua
DifferentialOrange: Changed

metrics/plugins/json.lua
DifferentialOrange: Reworked

metrics/plugins/json.lua
DifferentialOrange: There is no more gsub

metrics/plugins/json.lua
Totktonada: I still don't understand why `tonumber()` is there. It'll break some numbers outside of the [-2^53; 2^53] range:

```lua
tarantool> tonumber(9007199254740993ULL)
---
- 9007199254740992
...

tarantool> tonumber64(9007199254740993ULL)
---
- 9007199254740993
...
```

test/plugins/json_test.lua
Totktonada: `51589ULL` is less than 10^14 (where we started to encode it into cdata) and less than 2^53, where we can loss precision below 1.0. We should check that, say `9007199254740992ULL` (2^53) and numbers around it are marshalled correctly.

You can consider this following illustration how integral numbers are marshalled in tarantool:

https://github.com/tarantool/conf/blob/80078f85423ffd54d812c11b47aa85729e9e3d87/conf/client/etcd/scalar_serializer.lua#L134-L152

metrics/plugins/json.lua
DifferentialOrange: Well, dodging precision loss is impossible until we support encoding cdata to numbers in json package. Metric value as string is not supported by Telegraf/InfluxDB, who is the main (if not the only one) consumer of json metrics.

test/plugins/json_test.lua
DifferentialOrange: Looks like space bsize is always cdata
```
Tarantool 2.10.0-beta1-0-g7da4b1438
type 'help' for interactive help
tarantool> box.cfg{}
---
...
tarantool> box.space._space:bsize()
---
- 4151
...

tarantool> type(box.space._space:bsize())
---
- cdata
...
```

Totktonada: (Tarantool 2.10.0-beta1-91-g16a484a30.)

```
tarantool> print(json.encode(box.space._space:bsize()))
7943
---
...
```

Do you see another behaviour of the json encoder on another tarantool version?

metrics/plugins/json.lua
DifferentialOrange: Seems that I am wrong about cdata transform.

test/plugins/json_test.lua
DifferentialOrange: Actually no, it is as you described. Looks like I was wrong. Thank you! I had reworked json export to be as simple as possible.

metrics/plugins/json.lua
Totktonada: Okay.

test/plugins/json_test.lua
Totktonada: Okay so.

Totktonada:
I don't know, whether you expect a review from me. Actually I never work with metrics code, so I would prefer to don't review. But all my comments are resolved now.

DifferentialOrange:
@yngvar-antonsson , can you review this one?

yngvar-antonsson (approved):

Issue #235, opened 2021-05-12
Seems quantile overflow with undefined behaivior


filonenko-mikhail:
```lua
quantile = require('metrics.quantile')
fiber = require('fiber')

q = quantile.NewTargeted({[0.5]=0.01, [0.9]=0.01, [0.99]=0.01})
print(q)

for i=1,1e6 do quantile.Insert(q, math.random(1)) end

for i=1,200 do 
    fiber.new(function()
        for i=1,1e2 do quantile.Insert(q, math.random(1000)) end
    end)
end

for i=1,1e6 do quantile.Insert(q, math.random(1000)) end

require('console').start()
```

```
LuajitError: ../test2.lua:11: attempt to call field 'Insert' (a nil value)
Segmentation fault
  code: 128
  addr: (nil)
  context: 0x7f64519ff780
  siginfo: 0x7f64519ff8b0
  rax      0x40173800         1075263488
  rbx      0x4165c010         1097187344
  rcx      0x4                4
  rdx      0x1800             6144
  rsi      0x4063a00000000000 4639727962982187008
  rdi      0x0                0
  rsp      0x7f64519ffd30     140068842896688
  rbp      0x0                0
  r8       0x4063a00000000000 4639727962982187008
  r9       0x633aa0           6503072
  r10      0x414ebf30         1095679792
  r11      0x414fe470         1095754864
  r12      0xffffffff         4294967295
  r13      0x22               34
  r14      0x4165c3b8         1097188280
  r15      0x417b91f0         1098617328
  rip      0x633266           6500966
  eflags   0x10206            66054
  cs       0x33               51
  gs       0x0                0
  fs       0x0                0
  cr2      0x0                0
  err      0x0                0
  oldmask  0x0                0
  trapno   0xd                13
Current time: 1620824508
Please file a bug at http://github.com/tarantool/tarantool/issues
Attempting backtrace... Note: since the server has already crashed, 
this may fail as well
#0  0x5e9f19 in print_backtrace+9
#1  0x455ff7 in _ZL12sig_fatal_cbiP9siginfo_tPv+c7
#2  0x7f645a6e8630 in _L_unlock_13+34
#3  0x633266 in lj_alloc_free+326
#4  0x641700 in lj_cdata_free+80
#5  0x613c90 in gc_sweep+a0
#6  0x6146d4 in gc_onestep+44
#7  0x614e64 in lj_gc_step+54
#8  0x627ae8 in lj_cf_ffi_meta___index+b8
[...]

PR #332, opened 2021-12-08
Add upper constraint to quicksort

yngvar-antonsson:
Added to ensure that #235 won't come back

I didn't forget about

- [x] Tests


sergos (approved):
LGTM.

At first I thought the assert can be changed for update of the `low` and `high`. But it will not help to find the bug why appeared beyond the limits - the interface is internal.

PR #335, opened 2021-12-27
Log incorrect result from http_handler

### LINKED ISSUE #328: Error message from callback is lost in http_middleware
I have some error callback

```lua
metrics.register_callback(function() error('Callback failed') end)
```

And cartridge `app.roles.metrics` role for http endpoints.

When I try to collect metrics over HTTP I get the same response and the same log:

```lua
{
    response = {
        body = "Unhandled error: incorrect http handler for GET /eapi/rtm-customer-profile-app/manage/metrics: expecting return response object\
stack traceback:\
\9...tomer-profile-app/.rocks/share/tarantool/http/server.lua:743: in function 'process_client'\
\9...tomer-profile-app/.rocks/share/tarantool/http/server.lua:1199: in function <...tomer-profile-app/.rocks/share/tarantool/http/server.lua:1198>\
\9[C]: in function 'pcall'\
\9builtin/socket.lua:1081: in function <builtin/socket.lua:1079>\
\
\
\
Request:\
GET /eapi/rtm-customer-profile-app/manage/metrics? HTTP/1.1\13\
Host: localhost:8089\13\
Accept: */*\13\
Connection: close\13\
\13\
",
        headers = {
            connection = "close",
            ["content-length"] = "605",
            server = "Tarantool http (tarantool v2.8.2-0-gfc96d10f5-r428-macos)",
        },
        proto = {1, 1},
        reason = "Unknown",
        status = 500,
    },
    type = "HTTPRequest",
}
```

---
cartridge 2.7.1-1
metrics 0.11.0-1

yngvar-antonsson:
Close #328

### DIFF
--- a/metrics/http_middleware.lua
+++ b/metrics/http_middleware.lua
@@ -74,6 +74,7 @@ end
 function export.observe(collector, route, handler, ...)
     return collector:observe_latency(function(ok, result)
         if type(result) ~= 'table' then
+            log.error('http handler returned %q', result)
             error(('incorrect http handler for %s %s: expecting return response object'):
                 format(route.method, route.path), 0)
         end
--- a/metrics/http_middleware.lua
+++ b/metrics/http_middleware.lua
@@ -73,8 +73,10 @@ end
 -- @return value from observable function
 function export.observe(collector, route, handler, ...)
     return collector:observe_latency(function(ok, result)
-        if type(result) ~= 'table' then
+        if ok ~= true then
             log.error('http handler returned %q', result)
+        end
+        if type(result) ~= 'table' then
             error(('incorrect http handler for %s %s: expecting return response object'):
                 format(route.method, route.path), 0)
         end
--- a/test/http_middleware_test.lua
+++ b/test/http_middleware_test.lua
@@ -1,6 +1,7 @@
 local t = require('luatest')
 local g = t.group()
 
+local luatest_capture = require('luatest.capture')
 local utils = require('test.utils')
 
 local fun = require('fun')
@@ -160,6 +161,24 @@ g.test_v1_wrong_handler = function()
         'incorrect http handler for POST /some/path: expecting return response object',
         http_middleware.v1(handler, observer), request
     )
+end
+
+g.test_v1_handler_raise_an_error = function()
+    local request = {endpoint = table.copy(route)}
+    local observer = stub_observer(function() end)
+    local handler = function()
+        error('Handler is broken')
+    end
+
+    local capture = luatest_capture:new()
+
+    capture:wrap(true, function()
+        pcall(http_middleware.v1(handler, observer), request)
+    end)
+
+    local captured = capture:flush()
+
+    t.assert_str_contains(captured.stderr, 'Handler is broken')
 end
 
 g.test_v2_middleware = function()


metrics/http_middleware.lua
```diff
@@ -74,6 +74,7 @@ end
 function export.observe(collector, route, handler, ...)
     return collector:observe_latency(function(ok, result)
         if type(result) ~= 'table' then
+            log.error('http handler returned %q', result)
```
DifferentialOrange, line 77: If `ok == false`, `err` still could be a table.

metrics/http_middleware.lua
```diff
@@ -74,6 +74,7 @@ end
 function export.observe(collector, route, handler, ...)
     return collector:observe_latency(function(ok, result)
         if type(result) ~= 'table' then
```
DifferentialOrange, line 76: If `ok == false`, `err` still could be a table.

metrics/http_middleware.lua
```diff
@@ -73,6 +73,9 @@ end
 -- @return value from observable function
 function export.observe(collector, route, handler, ...)
     return collector:observe_latency(function(ok, result)
+        if ok ~= true then
+            log.error('http handler returned %q', result)
```
DifferentialOrange, line 77: You can test it with luatest capture.

I don't know if you have the access, but you can see example here: https://github.com/tarantool/megafon-customer-search-index-app/blob/525ca9d9528336a0568af76b43b467a237fc52a9/test/integration/bootstrap_test.lua#L470-L481

test/http_middleware_test.lua
```diff
@@ -162,6 +163,24 @@ g.test_v1_wrong_handler = function()
     )
 end
 
+g.test_v1_handler_raise_an_error = function()
+    local request = {endpoint = table.copy(route)}
+    local observer = stub_observer(function() end)
+    local handler = function()
+        error('Handler is broken')
```
DifferentialOrange, line 170: The original issue is about error thrown with callback https://github.com/tarantool/metrics/issues/328. Does PR cover it? Or the case mentioned in the issue is "/metrics handler wrapped in HTTP middleware"?

DifferentialOrange: Well, based on repo search, the only place where you can get "incorrect http handler" error is here, so it should fix the behavior. Error thrown on invoke callback is not pcalled anywhere but here.

DifferentialOrange (approved):
I added some comments, but looks like everything is ok and there are no more things to discuss about this PR.

But the issue raises another question: how we should treat callback error? Maybe it would be a good idea to call everything that works and ignore everything that thrown an error. It is a bit disappointing that having an error in a single callback forbids user from obtaining all the metrics (both set with ok callbacks and ones do not set with callbacks). Maybe #328 should be followed by another issue.

yngvar-antonsson:
Sorry, haven't seen your comments before review rerequest

yngvar-antonsson:
> It is a bit disappointing that having an error in a single callback forbids user from obtaining all the metrics (both set with ok callbacks and ones do not set with callbacks). Maybe #328 should be followed by another issue.

Now error in callback is treated like a 500-error. Or I don't understand the question?

PR #339, opened 2022-01-17
Deprecate rps metrics

yngvar-antonsson:
Those metrics are useless and could be transformed from the other metrics.

I didn't forget about

- [x] Changelog


DifferentialOrange (approved):
Those metrics not just "could", they should be transformed from other metrics to prevent data loss.

PR #358, opened 2022-04-26
perf: 100x increases speed for collectors/counter if used without label

ochaton:
jitable counter:inc and counter:set if no labels used

I didn't following:

- [ ] Tests
- [ ] Changelog
- [ ] Rockspec and rpm spec

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Float numbers in Graphite exporter
 - Signed timestamp in Graphite exporter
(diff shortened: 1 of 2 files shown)


filonenko-mikhail (approved):

ochaton:
# On master branch
```
tarantool> clock.bench(function() return fun.range(1, 1e7):map(function(t) counter:inc(1) return t end):length() end)
---
- - 3.367997
  - 10000000
...
```

# On perf/v.grubov branch

```
tarantool> clock.bench(function() return fun.range(1, 1e7):map(function(t) counter:inc(1) return t end):length() end)
---
- - 0.028591
  - 10000000
...
```

PR #367, opened 2022-05-30
Forbid observation of non-number value in collectors

### LINKED ISSUE #365: Metric value should be a number
Recently I have worked on TDG Grafana Dashboard. There are some cases (I'm researching it right now) when some metrics have not a numerical, but boolean value:
```
tdg_kafka_topic_partitions_desired{partition="0",client_id="rdkafka",alias="tdg",type="consumer",connector_name="kafka",name="rdkafka#consumer-1",topic="orders"} true
```

It breaks up Prometheus metrics collection:
image

The solution is obvious: just put appropriate values in collectors. However it's easy to miss when, for example, you're exporting metrics from external module:
```lua
local function update_partitions_metrics(name, stats, labels)
    if stats == nil then
        return
    end

    local partition_gauges = partitions_gauges[name]
    if partition_gauges == nil then
        return
    end

    for partition_name, partition_stats in pairs(stats) do
        labels = table.deepcopy(labels)
        labels.partition = partition_name

        for gauge_name, partition_gauge in pairs(partition_gauges) do
            if partition_stats[gauge_name] ~= nil then
                partition_gauge:set(partition_stats[gauge_name], labels)
            end
        end
    end
end
```

So we should do something with it. Maybe each collector should have something like `checks('number')` on observe. Maybe we should add some condition to strip invalid values on Prometheus export. Having such a simple way to break up metrics collection sounds not good.

yngvar-antonsson:
I didn't forget about

- [x] Tests
- [x] Changelog

Close #365

### DIFF
--- a/metrics/collectors/shared.lua
+++ b/metrics/collectors/shared.lua
@@ -60,6 +60,9 @@ function Shared:remove(label_pairs)
 end
 
 function Shared:set(num, label_pairs)
+    if num ~= nil and type(tonumber(num)) ~= 'number' then
+        log.error("Collector set value should be a number. In the next release it will raise an error")
+    end
     num = num or 0
     local key = self.make_key(label_pairs)
     self.observations[key] = num
@@ -67,6 +70,9 @@ function Shared:set(num, label_pairs)
 end
 
 function Shared:inc(num, label_pairs)
+    if num ~= nil and type(tonumber(num)) ~= 'number' then
+        error("Collector increment should be a number")
+    end
     num = num or 1
     local key = self.make_key(label_pairs)
     local old_value = self.observations[key] or 0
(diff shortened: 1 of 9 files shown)


metrics/collectors/shared.lua
```diff
@@ -60,13 +60,19 @@ function Shared:remove(label_pairs)
...
+        error("Collector set value should be a number")
+    end
     num = num or 0
     local key = self.make_key(label_pairs)
     self.observations[key] = num
     self.label_pairs[key] = label_pairs or {}
 end
 
 function Shared:inc(num, label_pairs)
+    if num and type(tonumber(num)) ~= 'number' then
```
DifferentialOrange, line 73: I prefer explicit `num == nil` cause it's more readable, covers `box.NULL` and treats `num == false` case as should.

metrics/collectors/summary.lua
```diff
@@ -54,6 +54,9 @@ function Summary:observe(num, label_pairs)
     if label_pairs.quantile then
         error('Label "quantile" are not allowed in summary')
     end
+    if num and type(tonumber(num)) ~= 'number' then
```
DifferentialOrange, line 57: If I understand the code, `num == nil` is invalid here too.

DifferentialOrange (commented):
To be honest, I'm not sure that this breaking change will not make things worse, but it's up to you and your team to decide the package roadmap and verify things with product managers.

DifferentialOrange (approved):

PR #377, opened 2022-06-21
remove old thread info from cpu metrics

olegrok:
In case if application starts new threads it's possible that in
result thread info will contain info about all threads that were
started ever. As result size of such metrics will be quite huge.
This patch removes info about dead threads and user will see info
only about relevant threads.

### DIFF
--- a/metrics/psutils/cpu.lua
+++ b/metrics/psutils/cpu.lua
@@ -12,25 +12,42 @@ local instance_file = arg[0]
 collectors_list.cpu_count = utils.set_gauge('cpu_count', 'The number of processors', psutils.get_cpu_count())
 collectors_list.cpu_number = utils.set_gauge('cpu_number', 'The number of processors', psutils.get_cpu_count())
 
+local threads = {}
+
 local function update_cpu_metrics()
     utils.set_gauge('cpu_total', 'Host CPU time', psutils.get_cpu_time())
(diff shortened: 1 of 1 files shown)


yngvar-antonsson:
Please add Changelog and test (if it's possible)

yngvar-antonsson (approved):

metrics/psutils/cpu.lua
```diff
@@ -12,25 +12,42 @@ local instance_file = arg[0]
...
+        utime_labels.kind = 'user'
+        collectors_list.cpu_thread = utils.set_gauge('cpu_thread', 'Tarantool thread cpu time',
+            thread_info.utime, utime_labels)
+
+        local stime_labels = table.copy(labels)
+        stime_labels.kind = 'user'
+        collectors_list.cpu_thread = utils.set_gauge('cpu_thread', 'Tarantool thread cpu time',
+            thread_info.stime, stime_labels)
+
+        threads[thread_info.pid] = nil
```
opomuc, line 39:
```
tarantool> a = {}
---
...

tarantool> a[1000]=nil
---
...

tarantool> a
---
- []
...

tarantool> a[100033]=nil
---
...

tarantool> a
---
- []
...

tarantool> a['100033']=nil
---
...

tarantool> a
---
- []
...

tarantool>
```

я чего-то не понимаю?

opomuc (changes requested):

opomuc (approved):

PR #416, opened 2022-12-06
Fix metrics version

### LINKED ISSUE #330: Provide a way to check version in code
`metrics` is a developing package and new features or critical bugfixes happen often, as well as new releases. There are no explicit way to check what metrics version is used in code when you are providing an optional integration with your module. For example, the only way to check that I'm using metrics `0.9.0` or greater with age buckets and fixed summary under high load is to check `require('metrics.collectors.summary').rotate_age_buckets ~= nil`, but this check is based on internal structure of the module that could change at any time. It would be nice to have something like `require('cartridge.VERSION')`.

yngvar-antonsson:
Close #330

### DIFF
--- a/.github/workflows/push_rockspec.yml
+++ b/.github/workflows/push_rockspec.yml
@@ -33,15 +33,13 @@ jobs:
 
       # https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions
       - name: Set env
-        run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+        run: echo "GIT_TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
 
       - name: Push release rockspec
         run: |
-          sed \
-            -e "s/branch = '.\+'/tag = '${GIT_TAG}'/g" \
-            -e "s/version = '.\+'/version = '${GIT_TAG}-1'/g" \
-            ${{ env.ROCK_NAME }}-scm-1.rockspec > ${{ env.ROCK_NAME }}-${GIT_TAG}-1.rockspec
-          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${GIT_TAG}"-1'/g" metrics/init.lua
+          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${env.GIT_TAG}"-1'/g" metrics/init.lua
+      - run: tarantoolctl rocks install ${{ env.ROCK_NAME }}-${{ env.GIT_TAG }}-1.rockspec
+      - run: tarantoolctl rocks pack ${{ env.ROCK_NAME }} ${{ env.GIT_TAG }}
 
       - uses: tarantool/rocks.tarantool.org/github-action@master
         with:


.github/workflows/push_rockspec.yml
```diff
@@ -33,15 +33,13 @@ jobs:
 
       # https://stackoverflow.com/questions/58177786/get-the-current-pushed-tag-in-github-actions
       - name: Set env
-        run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+        run: echo "GIT_TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
```
Totktonada, line 36: According to the [documentation](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables) (and the `startWith()` operator above) the variable is like `refs/tags/X.Y.Z`. So both variants should work:

Bash:

```shell
$ (GITHUB_REF=refs/tags/X.Y.Z; echo ${GITHUB_REF#refs/*/})
X.Y.Z
$ (GITHUB_REF=refs/tags/X.Y.Z; echo ${GITHUB_REF##*/})
X.Y.Z
```

Dash:

```shell
$ (GITHUB_REF=refs/tags/X.Y.Z; echo ${GITHUB_REF#refs/*/})
X.Y.Z
$ (GITHUB_REF=refs/tags/X.Y.Z; echo ${GITHUB_REF##*/})
X.Y.Z
```

So what is the motivation of the change? To follow usual pattern from other modules? Please, clarify it somewhere (commit message, pull request description).

.github/workflows/push_rockspec.yml
```diff
@@ -33,15 +33,13 @@ jobs:
...
 
       - name: Push release rockspec
         run: |
-          sed \
-            -e "s/branch = '.\+'/tag = '${GIT_TAG}'/g" \
-            -e "s/version = '.\+'/version = '${GIT_TAG}-1'/g" \
-            ${{ env.ROCK_NAME }}-scm-1.rockspec > ${{ env.ROCK_NAME }}-${GIT_TAG}-1.rockspec
-          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${GIT_TAG}"-1'/g" metrics/init.lua
+          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${env.GIT_TAG}"-1'/g" metrics/init.lua
```
Totktonada, line 40: In tarantool/graphql I decided to get rid of the rockspec revision. You can look at motivation and implementation here: https://github.com/tarantool/graphql/commit/0b4ec593a9677d010ec52d9c5b49852050884155.

.github/workflows/push_rockspec.yml
```diff
@@ -33,15 +33,13 @@ jobs:
...
       - name: Set env
-        run: echo "GIT_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+        run: echo "GIT_TAG=${GITHUB_REF##*/}" >> $GITHUB_ENV
 
       - name: Push release rockspec
         run: |
-          sed \
-            -e "s/branch = '.\+'/tag = '${GIT_TAG}'/g" \
-            -e "s/version = '.\+'/version = '${GIT_TAG}-1'/g" \
-            ${{ env.ROCK_NAME }}-scm-1.rockspec > ${{ env.ROCK_NAME }}-${GIT_TAG}-1.rockspec
```
Totktonada, line 43: Further commands use this generated file. They will fail if it doesn't exists.

.github/workflows/push_rockspec.yml
```diff
@@ -33,15 +33,13 @@ jobs:
...
-          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${GIT_TAG}"-1'/g" metrics/init.lua
+          sed -i "s/local VERSION = 'scm-1'/local VERSION = '"${env.GIT_TAG}"-1'/g" metrics/init.lua
+      - run: tarantoolctl rocks install ${{ env.ROCK_NAME }}-${{ env.GIT_TAG }}-1.rockspec
+      - run: tarantoolctl rocks pack ${{ env.ROCK_NAME }} ${{ env.GIT_TAG }}
```
Totktonada, line 42: Here we create an `.all.rock` file and **don't** upload it. So the module with patched version string will not present on the rock server.

I'm not sure that `rocks install` will install the module with patched version string. I guess it'll do a fresh clone of the repository pointed by `source.{url,branch,tag}`. Did you check it? I use `rocks make` for this purpose. You can find background information here: https://github.com/tarantool/graphql/commit/029fcf2f0475024db8629e270d613a2da12b2a4b.

PR #419, opened 2022-12-12
Fix get_nprocs_conf symbol not found error on MacOS

yngvar-antonsson:


metrics/psutils/psutils_linux.lua
```diff
@@ -65,5 +69,5 @@ end
 return {
     get_cpu_time = get_cpu_time,
     get_process_cpu_time = get_process_cpu_time,
-    get_cpu_count = ffi.C.get_nprocs_conf,
+    get_cpu_count = get_nprocs_conf,
```
filonenko-mikhail, line 72: is it nil for macos. will it make an error atttempt to call nil value somewhere in metrics code.

yngvar-antonsson: That code won't be called on Macos:

https://github.com/tarantool/metrics/blob/4992574588687a1037c3d3b3bcf6dd910d1a3245/metrics/psutils/cpu.lua#L2-L4

PR #428, opened 2023-01-27
Fix sed command

yngvar-antonsson:


.github/workflows/push_rockspec.yml
```diff
@@ -38,8 +38,8 @@ jobs:
       - name: Push release rockspec
         run: |
           sed \
-            -e "s/branch = '.\+'/tag = '${GIT_TAG}'/g" \
-            -e "s/version = '.\+'/version = '${GIT_TAG}-1'/g" \
+            -e "s/branch = \".\+\"/tag = '${GIT_TAG}'/g" \
+            -e "s/version = \".\+\"/version = '${GIT_TAG}-1'/g" \
```
Totktonada, line 42: `branch` and `version` are with single quotes in https://github.com/tarantool/metrics/blob/master/metrics-scm-1.rockspec.

yngvar-antonsson: Oh, thank you

PR #434, opened 2023-02-01
internal: remove circular dependencies

### LINKED ISSUE #433: Redesign dependency structure
Part of tarantool/tarantool#7725

It first started when I was needed to define the order of module files to load. And there were several unexpected surprises.

image

To begin with, having requires inside functions already seems like a smell of a bad dependency structure.

What's external and what's internal for the package?

External:
- `metrics`,
- `metrics.plugins.*`,
- `cartridge.roles.metrics`,
- `metrics.psutils.cpu`.

Users require `metrics` to work with metrics, metrics plugins to export metrics, cartridge role to work with metrics from cartridge. The latter one is a weird one -- it's something like an old default metrics, so we won't discuss it.

It makes sense for plugins and cartridge role to depend on main metrics package since it's a wrapper. What makes less sense is that more than half of the internal package scripts (`metrics.tarantool.*` and others) depend on the main metrics package. They are not expected to be used directly be the user, only through the main package (yeah, it is as confusing as it sounds). So the questionable places are

image

and

image

and all in-function requires (the only one not covered by the two schemes above is a weird require of `metrics.collector.*` in HTTP middleware).

Such structure is also a block to some planned tarantool/tarantool#7725 features (specifically, `metrics.cfg` reload support) and it makes introducing new features (like whole `metrics.cfg`) more confusing.

DifferentialOrange:
This PR removes existing circular dependencies in module, as well as remaining `require`s inside the functions.

Before:

image

After:

image

(Still seems confusing, but at least more hierarchic.)

It is a prerequirement to `metrics.cfg{}` reload support.

I didn't forget about

- Tests (still green)
- Changelog (internal)
- Documentation (README and rst) (internal)
- [x] Rockspec and rpm spec

Closes #433, part of tarantool/tarantool#7725

### DIFF
--- a/metrics-scm-1.rockspec
+++ b/metrics-scm-1.rockspec
@@ -23,6 +23,7 @@ build = {
 
     modules = {
         ['metrics']                         = 'metrics/init.lua',
+        ['metrics.api']                     = 'metrics/api.lua',
         ['metrics.registry']                = 'metrics/registry.lua',
         ['metrics.quantile']                = 'metrics/quantile.lua',
         ['metrics.http_middleware']         = 'metrics/http_middleware.lua',
--- a/metrics/api.lua
+++ b/metrics/api.lua
@@ -1,0 +1,148 @@
+-- vim: ts=4:sw=4:sts=4:expandtab
+
+local checks = require('checks')
+
+local Registry = require('metrics.registry')
+
+local Counter = require('metrics.collectors.counter')
+local Gauge = require('metrics.collectors.gauge')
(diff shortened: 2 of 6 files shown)


yngvar-antonsson:
Is this schema handmade or generated?

yngvar-antonsson (approved):

DifferentialOrange:
> Is this schema handmade or generated?

I didn't even though that it is possible to automatize the process, so it's hand-made with app.diagrams.net . (I can share the sources if you need them.)

PR #440, opened 2023-03-14
Override builtin

DifferentialOrange:
The ability to override built-in modules was introduced with [1]. To use
it, you must install the package files using the override.package path
for each file included.

Using soft links inside the rock seems like a bad idea -- luarocks try
to resolve them on files deploy and fails to do so. The only stable
approach is seems to be to duplicate the files to the override
folder.

builtin rockspec doesn't allow to specify a single file twice, so we
need to store the soft link in the repo to workaround it. To build
override package with builtin rockspec, we'll also need to manually
maintain two lists of package files.

cmake approach allows to automatize it. On the other hand, to work with
cmake-built rock, you'll need to `make` the rock while working locally.
Since `make .rocks` is a dependency of `make test`, it shouldn't be
a big issue.

After [2], metrics will be embedded to core Tarantool. The ability to
override them with installed rock will make it possible to use old
Tarantool with new metrics. cartridge role, which was embedded to
tarantool/cartridge in [3], is also would be override.

1. https://github.com/tarantool/tarantool/issues/7774
2. https://github.com/tarantool/tarantool/issues/7725
3. https://github.com/tarantool/cartridge/pull/2047

Closes tarantool/tarantool#7727


metrics-scm-1.rockspec
```diff
@@ -19,51 +19,10 @@ dependencies = {
...
-        ['metrics.tarantool.luajit']        = 'metrics/tarantool/luajit.lua',
-        ['metrics.tarantool.vinyl']         = 'metrics/tarantool/vinyl.lua',
-        ['metrics.utils']                   = 'metrics/utils.lua',
-        ['metrics.cfg']                     = 'metrics/cfg.lua',
-        ['metrics.stash']                   = 'metrics/stash.lua',
-        ['metrics.version']                 = 'metrics/version.lua',
-        ['cartridge.roles.metrics']         = 'cartridge/roles/metrics.lua',
-        ['cartridge.health']                = 'cartridge/health.lua',
-    }
+    type = 'cmake',
```
Totktonada, line 22: > cmake allows more flexible setup, which is needed for automatic override
package build.

What exactly doesn't work with `type = 'builtin'`? I guess that

```lua
modules = {
    ['metrics'] = 'metrics/init.lua',
    ['metrics.api'] = 'metrics/api.lua',
    <...>
    ['override.metrics'] = 'metrics/init.lua',
    ['override.metrics.api'] = 'metrics/api.lua',
    <...>
}
```

would work.

Totktonada: Georgy said that it actually doesn't work: each file is installed once (into override/) in the case.

He also shared a way to overcome this problem:

```sh
$ mkdir override
$ ln -s ../metrics override/metrics
```

And in the rockspec:

```lua
modules = {
    ['metrics'] = 'metrics/init.lua',
    ['metrics.api'] = 'metrics/api.lua',
    <...>
    ['override.metrics'] = 'override/metrics/init.lua',
    ['override.metrics.api'] = 'override/metrics/api.lua',
    <...>
}
```

However I agree that it looks like a crutch.

OTOH, it enables the overriding of the built-in module, when one just steps into the repository (it may be convenient at development).

I would say that it is up to the maintener. There are no strict cons or pros.

I would prefer if the commit message will contain more specific motivation behind the builtin -> cmake transition. So we can return here back in a future and understand, say, whether others should do exactly same or it depends.

DifferentialOrange:
Update: override cartridge role too

metrics-scm-1.rockspec
DifferentialOrange: > I would say that it is up to the maintener. There are no strict cons or pros.

I had opened an alternative PR https://github.com/tarantool/metrics/pull/444 . I'll discuss with current maintainer which solution is preferred

PR #444, opened 2023-03-23
Override builtin alternative

DifferentialOrange:
Alternative to #440 

The ability to override built-in modules was introduced with [1]. To use it, you must install the package files using the override.package path for each file included.

Using soft links inside the rock seems like a bad idea -- luarocks try to resolve them on files deploy and fails to do so. The only stable approach is seems to be to duplicate the files to the override folder.

builtin rockspec doesn't allow to specify a single file twice, so we need to store the soft link in the repo to workaround it. To build override package with builtin rockspec, we'll also need to manually maintain two lists of package files. On the other hand, soft link in repo allows to work with files locally ever without package rebuild.

After [2], metrics will be embedded to core Tarantool. The ability to override them with installed rock will make it possible to use old Tarantool with new metrics. cartridge role, which was embedded to tarantool/cartridge in [3], is also would be override.

1. https://github.com/tarantool/tarantool/issues/7774
2. https://github.com/tarantool/tarantool/issues/7725
3. https://github.com/tarantool/cartridge/pull/2047

Closes tarantool/tarantool#7727


DifferentialOrange:
We decided to stick to #440

PR #453, opened 2023-05-18
Move cartridge role to a separate repo

yngvar-antonsson:


.github/workflows/test.yml
```diff
@@ -91,8 +91,6 @@ jobs:
           go-version: '1.15'
 
       - name: promtool test
-        env:
```
DifferentialOrange, line 94: I think we should restore this to validate cartridge metrics names (like cartridge issues)

DifferentialOrange (approved):
Thank you for your PR. Seems fine, except for the last comment

PR #474, opened 2023-11-03
add an efficient labels serializer

ftelnov:
What has been done? Why? What problem is being solved?

I found out that observe method works slow. Further investigations showed that `make_key` function sorts keys every run time. `make_key` is being called often during, for instance, hist:observe. To optimize this case and both keep the default behavior, I propose two things:

1) Allow user to override `labels_pairs` serialization scheme with `__metrics_make_key` meta-method. If it is supplied, it would be used, otherwise the default sorting scheme would be used;
2) Provide precaching labels serializer. It would order the labels properly on the client side, and only concat pairs on the metrics side. The complexity is linear then.

I didn't forget about

- [x] Tests
- [x] Changelog
- [ ] Documentation (README and rst) (don't think I should alter it for such a small api?)
- [ ] Rockspec and rpm spec (I don't really know what to add here)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `metrics.cfg{}` `"all"` metasection for array `include` and `exclude`
   (`metrics.cfg{include={'all'}}` can be used instead of `metrics.cfg{include='all'}`,
   `metrics.cfg{exclude={'all'}}` can be used instead of `metrics.cfg{include='none'}`)
+- Allow users to decide on labels serialization scheme themselves. Add `labels_serializer`, which provides alternative efficient labels serialization scheme.
 
 ## [1.0.0] - 2023-05-22
 ### Changed
--- a/metrics/api.lua
+++ b/metrics/api.lua
@@ -147,7 +147,7 @@ end
 local function labels_serializer(labels_keys)
     table.sort(labels_keys)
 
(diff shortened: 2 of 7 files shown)


DifferentialOrange (commented):
Thank you for your PR! I have several question regarding implementation, testing and documentation. My main concern is how it will work with global labels -- for now it seems that one would need to explicilty support them in serializer, which, for example, will break the usage HTTP middleware after your change.

Since we are talking about efficiency here, it is also crucial to provide some performance comparison here. For example, I had used the following script to check your updates performance:
```lua
local clock = require('clock')

metrics.cfg{include = 'none', labels = {alias = 'myalias'}}

local custom_serializer = metrics.labels_serializer({'alias', 'l1', 'l2'})

local collector_with_old_serialization = metrics.counter('collector_with_old_serialization')
local bench_old = clock.bench(function()
    for i = 1, 1e6 do
        collector_with_old_serialization:inc(1, {l1 = 'v1', l2 = 'v' .. tostring(i)})
    end
end)

local collector_with_new_serialization = metrics.counter('collector_with_new_serialization')
local bench_new = clock.bench(function()
    for i = 1, 1e6 do
        collector_with_new_serialization:inc(1, custom_serializer.wrap{l1 = 'v1', l2 = 'v' .. tostring(i)})
    end
end)

print(('old: %.4f s, new: %.4f s'):format(bench_old[1], bench_new[1]))
```
```bash
old: 3.6006 s, new: 3.0873 s
```

PR #482, opened 2024-04-04
Log warning if histogram received cdata

### LINKED ISSUE #480: Incorrect handling of cdata values
Implicitly we add "inf" bucket to histogram:

https://github.com/tarantool/metrics/blob/d239e851471d3e98a9b23b6949c140c6a346b031/metrics/collectors/histogram.lua#L27-L29

Then we allow to observe any value that can be casted to number: numbers, strings (like "1", "2", "3") and ULL/LL cdata.

But this condition doesn't consider value type anyhow:
https://github.com/tarantool/metrics/blob/d239e851471d3e98a9b23b6949c140c6a346b031/metrics/collectors/histogram.lua#L58

In case of string it just throws:
```
tarantool> mutation_time:observe('1')
---
- error: '.../share/tarantool/override/metrics/collectors/counter.lua:9: attempt to
    compare string with number'
...
```

In case of cdata in incorrectly compares input value with `inf`:
```
tarantool> mutation_time:observe(1ULL)
-- input - bucket - [input < bucket]
1ULL	100	true
1ULL	200	true
1ULL	300	true
1ULL	400	true
1ULL	500	true
1ULL	600	true
1ULL	700	true
1ULL	800	true
1ULL	900	true
1ULL	1000	true
1ULL	inf	false
```

It's caused by LuaJIT semantics:
```
tarantool> inf = 1/0
---
...

tarantool> inf > 1
---
- true
...

tarantool> inf > 1ULL
---
- false
...

tarantool> inf > 1LL
---
- false
...
tarantool> tonumber('inf') == 0ULL
---
- true
...
```

It's very hard to find an issue with cdata numbers. I suggest to throw in such cases or signal about an issue somehow.

yngvar-antonsson:
Using cdata values as historgam observations is dangerous, as described in #480. On the other hand, plenty of clients (e.g. tdg) use cdata with this collector. We decided not to throw an error here, at least for now.  

I didn't forget about

- [x] Changelog

Close #480

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `tnt_election_leader_idle` metric.
 
+- Histogram now logs a warning if `observe` is called with `cdata` value.
+
+### Deprecated
+
+- Using `cdata` values with `histogram:observe`.
+
 ## [1.0.0] - 2023-05-22
 ### Changed
 
--- a/metrics/collectors/histogram.lua
+++ b/metrics/collectors/histogram.lua
@@ -52,7 +52,7 @@ function Histogram:observe(num, label_pairs)
         error("Histogram observation should be a number")
     end
     if not cdata_warning_logged and type(num) == 'cdata' then
-        log.warn("Using cdata as observation in historgam " ..
+        log.error("Using cdata as observation in historgam " ..
             "can lead to unexpected results. " ..
             "That log message will be an error in the future.")
         cdata_warning_logged = true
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - `tnt_election_leader_idle` metric.
 
+- Histogram now logs a warning if `observe` is called with `cdata` value.
```
olegrok, line 15: Let's call it "deprecated"

DifferentialOrange (approved):
It would be nice to have tests, but implementation seems fine

PR #491, opened 2024-07-04
Tarantool 3 "replacements" for Cartridge

DifferentialOrange:
Tarantool Grafana dashboard has some unique Cartridge panels. This PR is a part of "make Grafana dashboard work with Tarantool 3 with minimal efforts" activity (https://github.com/tarantool/grafana-dashboard/issues/224).

The state is as follows. Cartridge had the following metrics:
- tnt_cartridge_issues
- tnt_cartridge_failover_trigger_total
- tnt_clock_delta (from membership)

After this patch, `metrics` will have the following metrics for Tarantool 3:
- tnt_config_alerts
- tnt_config_status

How one covers the other? `tnt_config_alerts` is similar to `tnt_cartridge_issues`, but only for configuration apply. It covers everything that had gone wrong with applying configuration to the instance, separated to warnings and errors as well. It is a bit more detailed since Cartridge could just reject wrong configuration on two-phase commit and Tarantool 3 has asynchronous per-instance apply. `tnt_config_status` extend this info covering non-initialized instances and in-between states. On the other hand, config alerts does not cover other cluster issues, like "replication is broken due to conflict" and "instance 1 cannot ping instance 2 due to network issues". AFAIK, Tarantool 3 does not have any built-in mechanism for it. Also, Tarantool 3 (including newest EE failover) does not have any failover counters, so there is no analogues to `tnt_cartridge_failover_trigger_total` (yet existing `tnt_read_only` already provide a lot of information related to this question and also a part of Grafana dashboard for both Cartridge and Tarantool 3). There doesn't seem to be any replacement for `tnt_clock_delta` as well.

TCM also has some "replacement" for Cartridge issues, see [source code](https://github.com/tarantool/tarantool-cluster-manager/blob/master/server/internal/pkg/cluster/issues/issues.go). It covers two things: config alerts and vshard bootstrap success. We cover the first one. The second one has some additional assumptions and non-trivial logic, so I decided to not cover it here for now. It would be better if proper check would be added to `vshard` or `tarantool` so it can be exposed here, if required. So I think it's safe to say that here we cover almost the same things that are covered by TCM.

Visualization example:
image

- [x] Tests
- [x] Changelog
- Rockspec and rpm spec (not needed, already build metrics/*)

Part of https://github.com/tarantool/grafana-dashboard/issues/224

### DIFF
--- a/doc/monitoring/metrics_reference.rst
+++ b/doc/monitoring/metrics_reference.rst
@@ -993,3 +993,21 @@ Read view statistics
 
         *   -   ``tnt_memtx_index_read_view``
             -   Memory (in bytes) held for read views.
+
+
+Tarantool configuration
+-----------------------
+
+These metrics are available starting from Tarantool 3.0.
+
+..  container:: table
+
+    ..  list-table::
+        :widths: 25 75
+        :header-rows: 0
+
+        *   -   ``tnt_config_alerts``
+            -   Count of current instance :ref:`configuration apply alerts <config_api_reference_info>`.
+                ``{level="warn"}`` label covers warnings and
+                ``{level="error"}`` covers errors.
+                
--- a/doc/monitoring/metrics_reference.rst
+++ b/doc/monitoring/metrics_reference.rst
@@ -1010,4 +1010,23 @@ These metrics are available starting from Tarantool 3.0.
             -   Count of current instance :ref:`configuration apply alerts <config_api_reference_info>`.
                 ``{level="warn"}`` label covers warnings and
                 ``{level="error"}`` covers errors.
+
+        *   -   ``tnt_config_status``
+            -   The status of current instance :ref:`configuration apply <config_api_reference_info>`.
+                ``status`` label contains possible status name.
+                Current status has metric value ``1``, inactive statuses have metric value ``0``.
+
+                ..  code-block:: none
+
+                    # HELP tnt_config_status Tarantool 3 configuration status
+                    # TYPE tnt_config_status gauge
+                    tnt_config_status{status="reload_in_progress",alias="router-001-a"} 0
+                    tnt_config_status{status="uninitialized",alias="router-001-a"} 0
+                    tnt_config_status{status="check_warnings",alias="router-001-a"} 0
+                    tnt_config_status{status="ready",alias="router-001-a"} 1
+                    tnt_config_status{status="check_errors",alias="router-001-a"} 0
+                    tnt_config_status{status="startup_in_progress",alias="router-001-a"} 0
+
+                For example, this set of metrics means that current configuration
+                for ``router-001-a`` status is ``ready``.
                 
--- a/metrics/tarantool/config.lua
+++ b/metrics/tarantool/config.lua
@@ -1,0 +1,46 @@
+local utils = require('metrics.utils')
+
+local collectors_list = {}
+
+local function get_config_alerts(config_info)
+    -- https://github.com/tarantool/tarantool/blob/319357d5973d15d08b8eda6a230eada08b710802/src/box/lua/config/utils/aboard.lua#L17-L18
+    local config_alerts = {
+        warn = 0,
+        error = 0,
+    }
+
+    for _, alert in pairs(config_info.alerts) do
+        config_alerts[alert.type] = config_alerts[alert.type] + 1
+    end
+
+    return config_alerts
+end
+
+local function update()
+    if not utils.is_tarantool3() then
+        return
+    end
+
+    -- Can migrate to box.info().config later
+    -- https://github.com/tarantool/tarantool/commit/a1544d3bbc029c6fb2a148e580afe2b20e269b8d
+    local config = require('config')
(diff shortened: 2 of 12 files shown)


test/tarantool/cpu_metrics_test.lua
```diff
@@ -14,6 +14,8 @@ end)
 
 g.test_cpu = function(cg)
     cg.server:exec(function()
+        t.skip('Flaky test')
```
oleg-jukovec, line 17: It seems a good idea to create a new issue/add a link to the issue and fix the problem later.

oleg-jukovec (commented):
It looks like we will need two releases for the [grafana-dashboard](https://github.com/tarantool/grafana-dashboard/): without these metrics and with them.

Because we're currently aiming for Tarantool 3.1, and the metrics package will only be updated with Tarantool 3.2 in September.

DifferentialOrange:
> It looks like we will need two releases for the [grafana-dashboard](https://github.com/tarantool/grafana-dashboard/): without these metrics and with them.

Yeah, that's the plan. We already have 4 dashboards: Tarantool for Prometheus, Tarantool for InfluxDB, TDG for Prometheus, TDG for InfluxDB. After https://github.com/tarantool/grafana-dashboard/, there will be six of them. Due to grafana-dashboard jsonnet nature it's rather easy to build different dashboards (just a bit bothersome to support different release pages).

doc/monitoring/metrics_reference.rst
```diff
@@ -993,3 +993,40 @@ Read view statistics
...
+        :widths: 25 75
+        :header-rows: 0
+
+        *   -   ``tnt_config_alerts``
+            -   Count of current instance :ref:`configuration apply alerts <_config_api_reference_info>`.
+                ``{level="warn"}`` label covers warnings and
+                ``{level="error"}`` covers errors.
+
+        *   -   ``tnt_config_status``
+            -   The status of current instance :ref:`configuration apply <_config_api_reference_info>`.
```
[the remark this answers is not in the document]
DifferentialOrange: These are for Sphinx documentation build, they don't render on Github (and will render properly only as a part of tarantool/doc since they reference a base documentation path). On the other hand, you're still right and they were broken. Now everything is fine: 

image
(you can check here with proper credentials: https://docs.d.tarantool.io/en/doc/metrics-1.2.0/book/monitoring/metrics_reference/)

metrics/utils.lua
```diff
@@ -36,4 +36,21 @@ function utils.delete_collectors(list)
     table.clear(list)
 end
 
+-- https://github.com/tarantool/crud/blob/98b120ef7095fa34525ef9d335a1458a2edf0cca/crud/common/utils.lua#L604-L610
```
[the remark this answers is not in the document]
DifferentialOrange: Removed

test/tarantool/cpu_metrics_test.lua
DifferentialOrange: Added #492

DifferentialOrange:
>  and the metrics package will only be updated with Tarantool 3.2 in September
AFAIU, we will bump it for all supported 3.x series, so it will be built-in to 3.1.1 as well

oleg-jukovec (approved):
Thank you for the patch!

oleg-jukovec:
> AFAIU, we will bump it for all supported 3.x series, so it will be built-in to 3.1.1 as well

To be honest, I don't know anything about plans for 3.1.1

DifferentialOrange:
I have found one more cartridge-exclusive metrics: `tnt_clock_delta`. There doesn't seem to be any replacement in Tarantool 3 as well

(All other panels work fine.)

oleg-jukovec (approved):

metrics/tarantool/config.lua
```diff
@@ -0,0 +1,76 @@
...
+    local config_status = {
+        uninitialized = 0,
+        startup_in_progress = 0,
+        reload_in_progress = 0,
+        check_warnings = 0,
+        check_errors = 0,
+        ready = 0,
+    }
+
+    config_status[config_info.status] = 1
```
yngvar-antonsson, line 31: Maybe it's better to choose a unique numerical value for each status instead of passing the status name to labels?

DifferentialOrange: Such approach is bad at scaling: in case some new status will be introduced, we will have two ways:
- add a value in-between: very bad, will break existing dashboards;
- add a value to the tail: is bad in case it's not one more "healthy" status (which is unlikely), since healthy "ready" status will always be visualized as something between "reload_in_progress" and "new_not_healthy_status".

Such metrics are also impossible to read without documentation. Similar approach is already discussed in [1, 2]. So I think this approach is a bit better, even though it's harder to visualize.
1. https://github.com/prometheus/client_python/issues/416
2. https://github.com/open-telemetry/opentelemetry-specification/issues/1711

yngvar-antonsson (approved):

PR #505, opened 2025-04-29
Bump cmake version. Bump ubuntu version in the CI.

### LINKED ISSUE #504: Update workflows using Ubuntu 20.04 (focal)
Ubuntu 20.04 will soon be deprecated and unsupported in the GitHub infrastructure.

See also: https://github.com/actions/runner-images/issues/11101

So, we should update workflows, like reusable_testing.yml workflow which is used in the integraion Tarantool testing.
I suppose we can update it up to 24.04 (noble).

Needed for tarantool/tarantool#11220

Buristan:
ci: update ubuntu version

The `ubuntu-20.04 (focal)` image is no longer supported, so the build using this image returns the following error:

| This is a scheduled Ubuntu 20.04 brownout. Ubuntu 20.04 LTS runner
| will be removed on 2025-04-01.

See [1] for the details.

This patch bumps the used version to the ubuntu-24.04 (noble).

[1]: https://github.com/actions/runner-images/issues/11101

Closes #504
Needed for tarantool/tarantool#11220

---

ci: remove EOL Tarantool versions from testing

This patch drops EOL Tarantool versions that has no package available for Ubuntu 24.04 noble at download.tarantool.org.

---

cmake: update the minimum required version

For CMake 4.0, compatibility with versions of CMake older than 3.5 has been removed [2]. Thus, on macOS it leads to the error on build, since Homebrew updates it to version 4.0.

This patch sets the minimum required version to 3.5 for now.

[2]: https://cmake.org/cmake/help/latest/release/4.0.html#deprecated-and-removed-features


yngvar-antonsson (approved):

.github/workflows/test.yml
```diff
@@ -10,58 +10,7 @@ jobs:
...
-          - tarantool: "2.10"
-            cartridge: "2.10.0"
-          - tarantool: "2.10"
-            cartridge: "2.11.0"
-          - tarantool: "2.10"
-            cartridge: "2.12.4"
-          - tarantool: "2.10"
-            cartridge: "2.13.0"
-          - tarantool: "2.10"
-            cartridge: ""
```
yngvar-antonsson, line 64: I would add the latest cartridge versions here, but it's up to you

Buristan: For 2.10 there is no point in it, since there are no its packages for Ubuntu noble.

PR #518, opened 2025-09-09
cpu: add instance CPU usage metric

patapenka-alexey:
Closes #TNTP-4361


metrics/psutils/cpu.lua
```diff
@@ -12,6 +12,11 @@ local instance_file = arg[0]
...
 local threads = {}
 
+local instance_cpu_usage = {
+    system = 0,
+    user = 0
+}
+
 local function update_cpu_metrics()
     collectors_list.cpu_number = utils.set_gauge('cpu_number', 'The number of processors',
         psutils.get_cpu_count(), nil, nil, {default = true})
```
DifferentialOrange, line 22: It's cumulative, but stime and utime are cumulative as well. So it's the sum of the sum for some reason. I guess it should be

```suggestion
local instance_cpu_usage = {
    system = 0,
    user = 0
}

local function update_cpu_metrics()
    instance_cpu_usage.system = 0
    instance_cpu_usage.user = 0

    collectors_list.cpu_number = utils.set_gauge('cpu_number', 'The number of processors',
        psutils.get_cpu_count(), nil, nil, {default = true})
```

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,25 @@ local function update_cpu_metrics()
...
+    local instance_stime_labels = table.copy(cpu_label)
+    instance_stime_labels.kind = 'system'
+    collectors_list.instance_cpu_usage = utils.set_gauge('cpu_instance',
+        'Tarantool instance system CPU time', instance_cpu_usage.system,
+        instance_stime_labels, nil, {default = true})
+
+    local instance_utime_labels = table.copy(cpu_label)
+    instance_utime_labels.kind = 'user'
+    collectors_list.instance_cpu_usage = utils.set_gauge('cpu_instance',
+        'Tarantool instance user CPU time', instance_cpu_usage.user,
```
DifferentialOrange, line 63: The description belongs to a collector. You cannot set two different descriptions for the same collector with different observation labels.
```suggestion
        'Tarantool instance cpu time', instance_cpu_usage.user,
```

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,25 @@ local function update_cpu_metrics()
...
+        instance_cpu_usage.system = instance_cpu_usage.system + thread_info.stime
+        instance_cpu_usage.user = instance_cpu_usage.user + thread_info.utime
     end
 
+    local cpu_label = {}
+
+    local instance_stime_labels = table.copy(cpu_label)
+    instance_stime_labels.kind = 'system'
+    collectors_list.instance_cpu_usage = utils.set_gauge('cpu_instance',
+        'Tarantool instance system CPU time', instance_cpu_usage.system,
```
DifferentialOrange, line 57: Don't forget to update these as well.

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,25 @@ local function update_cpu_metrics()
...
         new_threads[thread_info.pid] = labels
+
+        instance_cpu_usage.system = instance_cpu_usage.system + thread_info.stime
+        instance_cpu_usage.user = instance_cpu_usage.user + thread_info.utime
     end
 
+    local cpu_label = {}
+
+    local instance_stime_labels = table.copy(cpu_label)
+    instance_stime_labels.kind = 'system'
```
DifferentialOrange, line 55: Why not just
```suggestion
    local instance_stime_labels = {kind = 'system'}
```

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,25 @@ local function update_cpu_metrics()
...
+    collectors_list.instance_cpu_usage = utils.set_gauge('cpu_instance',
+        'Tarantool instance system CPU time', instance_cpu_usage.system,
+        instance_stime_labels, nil, {default = true})
+
+    local instance_utime_labels = table.copy(cpu_label)
+    instance_utime_labels.kind = 'user'
```
DifferentialOrange, line 61: And here as well.

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,25 @@ local function update_cpu_metrics()
 
         threads[thread_info.pid] = nil
         new_threads[thread_info.pid] = labels
+
+        instance_cpu_usage.system = instance_cpu_usage.system + thread_info.stime
+        instance_cpu_usage.user = instance_cpu_usage.user + thread_info.utime
     end
 
+    local cpu_label = {}
```
DifferentialOrange, line 52: There's also `file_name` label below. You can include this one here to be consistent with `cpu_thread`. (And then `cpu_label` will make sense.) On the other hand, this one seems rather useless.

metrics/psutils/cpu.lua
patapenka-alexey: Updated.

metrics/psutils/cpu.lua
patapenka-alexey: Updated.

metrics/psutils/cpu.lua
patapenka-alexey: Updated.

metrics/psutils/cpu.lua
patapenka-alexey: Seems rather useless, but who am I to talk about this.
image

metrics/psutils/cpu.lua
```diff
@@ -39,8 +44,21 @@ local function update_cpu_metrics()
 
         threads[thread_info.pid] = nil
         new_threads[thread_info.pid] = labels
+
+        instance_cpu_usage.system = instance_cpu_usage.system + thread_info.stime
```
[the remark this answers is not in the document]
patapenka-alexey: and indeed...

PR #519, opened 2025-09-15
instance: total memory metric

patapenka-alexey:
This patch adds new `tnt_memory` metric.

Closes #TNTP-4363


doc/monitoring/metrics_reference.rst
```diff
@@ -35,7 +35,8 @@ The following metrics provide a picture of memory usage by the Tarantool process
     ..  list-table::
         :widths: 25 75
         :header-rows: 0
-
+        *   -   ``tnt_instance_memory``
```
DifferentialOrange, line 38: In the documentation, each section is enabled/disabled separately. If you introduce `instance` section, its metrics should be in a separate section.

DifferentialOrange: This part also should be updated

https://github.com/tarantool/metrics/blob/master/doc/monitoring/api_reference.rst#metrics-functions

metrics/tarantool/instance.lua
```diff
@@ -0,0 +1,76 @@
...
+        end
+        if memtx_stat_3.index ~= nil then
+            memory_stat = memory_stat + memtx_stat_3.index.total
+        end
+    end
+
+    --[[vinyl]]
+    local vinyl_stat = box.stat.vinyl()
+    if vinyl_stat ~= nil then
+        memory_stat = memory_stat + vinyl_stat.memory.tuple_cache
```
DifferentialOrange, line 54: It seems that `box.info.memory().cache` and `box.stat.vinyl().tuple_cache` are the same, so they're counted twice
> `cache` – the number of bytes used for caching user data. The memtx storage engine does not require a cache, so in fact this is the number of bytes in the cache for the tuples stored for the vinyl storage engine.
(https://www.tarantool.io/en/doc/latest/reference/reference_lua/box_info/memory/)

The same goes for `box.info.memory().data` and `box.info.memory().index`
> `data` – the number of bytes used for storing user data (the tuples) with the memtx engine and with level 0 of the vinyl engine, without taking memory fragmentation into account.
> `index` – the number of bytes used for indexing user data, including memtx and vinyl memory tree extents, the vinyl page index, and the vinyl bloom filters.

doc/monitoring/metrics_reference.rst
patapenka-alexey: Updated.

metrics/tarantool/instance.lua
patapenka-alexey: Updated.

doc/monitoring/metrics_reference.rst
```diff
@@ -25,6 +25,26 @@ General instance information:
...
+
+.. container:: table
+
+    ..  list-table::
+        :widths: 25 75
+        :header-rows: 0
+        *   -   ``tnt_cpu_user_time``
+            -   Tarantool CPU user time.
+        *   -   ``tnt_cpu_system_time``
+            -   Tarantool CPU system time.
```
oleg-jukovec, line 42: Is it a rebase artifact?

oleg-jukovec (changes requested):

oleg-jukovec (approved):

metrics/tarantool/memory.lua
```diff
@@ -15,6 +16,50 @@ local function update_memory_metrics()
                 nil, nil, {default = true})
         end
     end
+
+    local memory_stat = 0
+
+    if jit.os == 'Linux' then
+        local data = psutils.get_process_stats()
+        memory_stat = data.rss
```
Satbek, line 24: It’s confusing that this is the number of pages, not the actual memory size.

On macOS the metric will report bytes, while on Linux it reports pages.

If this goes into a dashboard, it could be misleading.

Maybe we should make two metrics: one for memory calculated through Tarantool calls, which will work on both macOS and Linux; and another one for RSS, which will work only on Linux.

PR #543, opened 2026-06-01
api: add selector-based filtering for custom metrics

mandesero:
This patch adds selector-based filtering for custom metrics.

Before this patch, the `metrics` module allowed applications to register custom collectors and callbacks, but there was no common way to group them and enable or disable them later through `metrics.cfg()`.

The patch introduces selectors for custom metrics. A selector may be attached to a collector or callback via metadata:

```lua

metrics.register_callback(function()
    in_flight:set(get_in_flight())
end, {selector = 'roles.crud-router'})
```

For convenience, metrics may also be registered through a namespace:

```lua

local crud = metrics.namespace('roles.crud-router')

crud:gauge('crud_requests'):set(1)
crud:gauge('crud_errors'):set(1)
```

The namespace prefixes registered collectors with hierarchical selectors:

```text
roles.crud-router.crud_requests
roles.crud-router.crud_errors
```

Selectors can be filtered with the new `metrics.set_filter()` API:

```lua
metrics.set_filter(
    {{selector = 'roles.crud-router'}},
    {{selector = 'roles.crud-router.crud_errors'}}
)
```

`include` enables matching selectors, `exclude` disables matching selectors, and `exclude` has priority. Matching is hierarchical, so `roles.crud-router` matches `roles.crud-router.crud_requests`.

`metrics.cfg()` now also accepts custom selector strings in `include` and `exclude`:

```lua
metrics.cfg{
    include = {'info', 'roles.crud-router'},
    exclude = {'roles.crud-router.crud_errors'},
}
```

Built-in Tarantool metric groups keep the existing behavior. Unknown strings in `include` / `exclude` are treated as custom selectors and are applied through the registry filter.

This lets applications and Tarantool roles expose custom metrics while still using the existing `metrics.include` / `metrics.exclude` configuration model.

Part of tarantool/tarantool#11358

### DIFF
--- a/metrics/api.lua
+++ b/metrics/api.lua
@@ -12,17 +12,24 @@ local Summary = require('metrics.collectors.summary')
 local registry = rawget(_G, '__metrics_registry')
 if not registry then
     registry = Registry.new()
+else
+    setmetatable(registry, Registry)
+    if registry.filter == nil then
+        registry:reset_filter()
+    end
 end
 registry.callbacks = {}
 
 rawset(_G, '__metrics_registry', registry)
 
 local function collectors()
-    return registry.collectors
+    return registry:filtered_collectors()
 end
 
-local function register_callback(...)
-    return registry:register_callback(...)
+local function register_callback(callback, metainfo)
+    checks('function', '?table')
+
+    return registry:register_callback(callback, metainfo)
 end
 
 local function unregister_callback(...)
@@ -48,7 +55,7 @@ local function collect(opts)
     end
 
     local result = {}
-    for _, collector in pairs(registry.collectors) do
+    for _, collector in pairs(registry:filtered_collectors()) do
         if opts.default_only then
             if collector.metainfo.default then
                 get_collector_values(collector, result)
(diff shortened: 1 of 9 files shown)


metrics/cfg.lua
```diff
@@ -7,6 +7,35 @@ local const = require('metrics.const')
 local stash = require('metrics.stash')
 local metrics_tarantool = require('metrics.tarantool')
 
+local function split_selectors(values, default_value, empty_default,
```
bigbes: Sorry for nitpicking, but i've spent quite some time trying to understand what `split_selectors` function means - worth adding description/renaming it.

metrics/registry.lua
```diff
@@ -70,4 +143,56 @@ function Registry:set_labels(label_pairs)
...
 end
 
+local function normalize_filter_side(value, default)
+    if value == nil then
+        return default
+    end
+
+    if type(value) == 'string' then
+        if value ~= 'all' and value ~= 'none' then
+            error('Metric selector must be a table with a selector field')
```
bigbes: misleading error message

metrics/registry.lua
```diff
@@ -70,4 +143,56 @@ function Registry:set_labels(label_pairs)
...
+        end
+        return {
+            all = value == 'all',
+            selectors = {},
+        }
+    end
+
+    if type(value) ~= 'table' then
+        error('Metric selector must be a table with a selector field')
```
bigbes: misleading error message

metrics/api.lua
```diff
@@ -12,17 +12,24 @@ local Summary = require('metrics.collectors.summary')
 local registry = rawget(_G, '__metrics_registry')
 if not registry then
     registry = Registry.new()
+else
```
bigbes, line 15: Do we have test-case for this? That looks like something that may be broken easily.

bigbes (changes requested):
Ty for you patch! Logic seems legit, but it's new functionality test coverage worth to be extended (for example - more cases for 'set_filter', reconfiguration, etc)

metrics/api.lua
mandesero: Added `metrics_test.test_hotreload_old_registry_gets_filter`

metrics/registry.lua
mandesero: Fixed.

metrics/registry.lua
mandesero: Fixed.

metrics/cfg.lua
mandesero: Replaced with `split_metric_groups_and_selectors` and added a corresponding comment.

patapenka-alexey (approved):
Thank you for the patch!

bigbes (approved):

PR #542, opened 2026-04-20
test: make memory metrics test portable

mandesero:
This patch makes `tarantool/memory_metrics_test.lua` portable across platforms.

Part of tarantool/tarantool#12559

PR #544, opened 2026-06-08
test: fix memory_metrics test for OSX

sergos:
The test is failing with local run on macbook.
The original problem is that psutils is present for Linux only.


patapenka-alexey (commented):
Hello! We've already [fixed](https://github.com/tarantool/metrics/pull/542/changes) this.

PR #549, opened 2026-07-01
test: disable config checks in config_metrics_test

Godrik0:
Set `config.checks` to `'off'` in `default_config` to avoid environment-dependent alerts (THP, readahead) in tests.

Part of tarantool/tarantool#12371

### DIFF
--- a/test/tarantool/config_metrics_test.lua
+++ b/test/tarantool/config_metrics_test.lua
@@ -59,6 +59,14 @@ local default_config =  {
     metrics = {
         include = {'all'},
     },
+    conditional = {
+        {
+            ['if'] = 'tarantool_version >= 3.8.1',
+            config = {
+                checks = 'off',
+            },
+        },
+    },
 }
 
 local function write_config(cg, config)


patapenka-alexey (approved):
Thank you for the patch!

oleg-jukovec (changes requested):
Thank you for the patch, but I’m lacking motivation - why do we need to do this, and why starting from version 3.8.0, not 3.4.0?

Please, discribe it in the PR/commit message.

elhimov (approved):

Godrik0:
@oleg-jukovec, This option disables production alerts in tests. Otherwise, tests that expect exact alert counts, such as those using `assert_config_alerts_metrics`, may fail when a CI runner has a configuration that triggers these alerts. For example, THP is enabled by default on some systems. The version is chosen based on when the feature becomes available in Tarantool CE.

oleg-jukovec:
> @oleg-jukovec, This option disables production alerts in tests. Otherwise, tests that expect exact alert counts, such as those using `assert_config_alerts_metrics`, may fail when a CI runner has a configuration that triggers these alerts. For example, THP is enabled by default on some systems. The version is chosen based on when the feature becomes available in Tarantool CE.

So why not to rewrite the test so that it takes this difference into account instead of disabling this functionality at all on the module level?

patapenka-alexey:
@Godrik0 as I understand, by disabling checks in the test config, we got the only alerts in the tests what the test itself injects, so the exact counts (warn=0/1, error=0/1) are stable on any runner. Without this change, on a machine where THP is enabled, Tarantool emits a warning, tnt_config_alerts{level="warn"} becomes 1 instead of the expected 0, and `test_config_metrics_if_healthy` fails in CI.
@oleg-jukovec we don't have any problems in this situation.

oleg-jukovec:
> @Godrik0 as I understand, by disabling checks in the test config, we got the only alerts in the tests what the test itself injects, so the exact counts (warn=0/1, error=0/1) are stable on any runner. Without this change, on a machine where THP is enabled, Tarantool emits a warning, tnt_config_alerts{level="warn"} becomes 1 instead of the expected 0, and `test_config_metrics_if_healthy` fails in CI. @oleg-jukovec we don't have any problems in this situation.

So let just rewrite the test:

```Lua
if THP_enabled then
    assert(len(warns) == 1)
else
    assert(len(warns) == 0)
end
```

Or is it impossible?

oleg-jukovec (approved):

PR #553, opened 2026-08-26
test: disable system alerts only if option exists

Totktonada:
This commit reimplements version-based check from PR #549 with a direct check of whether the given option (`config.checks`) exists. It gives a valid configuration for any tarantool 3.x version, including intermediate developer builds.

The motivation is that integration testing with metrics `master` fails in tarantool's CI, because the `config.checks` config option does not exist at the moment. While the feature that adds the option is under some discussion, let's just fix the CI.

Follows up PR #549
Related to tarantool/tarantool#12371


patapenka-alexey (approved):
Thank you for the patch!

test/tarantool/config_metrics_test.lua
```diff
@@ -18,56 +18,78 @@ g.after_all(function(cg)
...
+        if not config_lib.jsonschema then
+            return
+        end
+
+        local jsonschema = config_lib:jsonschema()
+        if jsonschema.properties.config.properties.checks == nil then
+            return
+        end
+
+        config.config = config.config or {}
```
elhimov: I'm not sure if it does matter or not but prior to this patch `checks = 'off'` was the part of `config.conditional.config` sub-table not `config.config`.

elhimov (commented):
LGTM except for one question

bigbes (approved):

patapenka-alexey:
Please, rebase your branch against the updated master.

test/tarantool/config_metrics_test.lua
Totktonada: The `conditional` section was to perform the version-based check, but I replaced it with the option-presence-check. So there is no more reason to keep `conditional`.

Totktonada:
Rebased on top of a fresh `master`.

test/tarantool/config_metrics_test.lua
patapenka-alexey: This is a tragic confluence of circumstances.

PR #555, opened 2026-08-26
ci: run CI on external pull requests

Totktonada:
Currently, CI is run only on pushing a branch. The jobs are shown in the pull request UI, when its source branch in the base repository. However, if I send a pull request from a fork, there are no CI checks.

Let's fix it similarly to how tarantool/crud does: run on branch push and pull request both, but skip the pull request run in jobs if the pull request is sent from the base repository. This way the jobs are not duplicated.

Also, while I'm here, add recent tarantool versions to the CI matrix.


patapenka-alexey (approved):
Thank you for the patch!

bigbes (approved):

